PR #6, opened 2026-06-18
birkhoff new instances (24, 32, 48, 64, 100)

victor-ibm:
This PR adds new instances for the Birkhoff problem.

The new problem sizes are 24, 32, 48, 64, and 100 (_sparse_ and _dense_). 

All files contain 10 instances, except for the 100 dense case. The 100-dense case has only 5 instances, since otherwise the file would be too large for GitHub.

**Note on naming:** The current filenames use the prefix qbench, which was the name we used before we chose qoblib. I suggest renaming the files to:

XXX_sparse.json / XXX_dense.json


mlschicker:
Hi @victor-ibm , 

thank you for providing the new instances!

Best, 
Max

PR #8, opened 2026-06-28
Add PDBO submission for LABS

liuwb17:
This submission adds PDBO results for LABS, covering instances 'labs002' through 'labs100'.
We are grateful to Prof. David Bernal (@bernalde) and Dr. Tim Bode (@timbode) for pointing out these challenging LABS instances to us.


mlschicker:
Hi @liuwb17 ,

thank you for the pull request. 
The results look very good - all solutions up to length 34 are optimal!

You mention that you run 1000 runs in parallel. 
Do you mean 1000 runs for the same instance in parallel?
Is the runtime averaged?
How do you define TTS in these runs?

Since your program is a heuristic, you could set the known optimal objective value (if it exists) as a cutoff, as there is no better solution to be found - this would drastically reduce your hardware-time.
I would be interested in the performance of this framework on some of the other problem classes like marketsplit and independent set. 
As your solver runs natively on multi-linear polynomials, it would be interesting to see if higher-order equivalent formulations for, e.g., independent set, solve differently. 

Have you thought about working directly on the spin domain?
It seems like that this would be a possibility with your present framework. 

A few things to fix before I can merge

- [x] Fill in the field `Coefficients Range` 
- [x] Fill in the field `# Non-Zero Coefficients` (Should be the number of terms in the multi-linear polynormial; this is interesting as it differs quite a lot when you work directly on the spin variables and don't substitute for binary)
- [x] Fill in the field `# Successful Runs` (Did all of your runs always reach the optimal solution?)
- [x] Run the solution checker with `--generate-readme` and push the READMEs.

Thank you!

Best, Max

liuwb17:
Hi Max,

I have addressed the four points you raised and resubmitted the results.

Regarding the runtime, PDBO optimizes 1000 primal candidates in parallel within a single GPU execution for the same instance. These are not 1000 independent repetitions with separate seeds. Therefore, I corrected the number of runs to 1. The reported total time is the wall-clock time of the run, and the TTS is the first timestamp at which PDBO reaches the reported objective value.

Thank you also for your constructive suggestion about trying the spin-domain formulation. PDBO was originally designed for binary variables, but it can indeed be naturally generalized to spin variables. In response, I have submitted the corresponding spin-domain results, denoted as PDBOspin. On the LABS instances, PDBOspin performs better than the binary-domain version overall.

Thank you!
Best, Liu

mlschicker:
Hi @liuwb17 ,

thank you for the update and clarifications.
The new results on the native spin domain are exciting!

Since you opened the PR, we have updated the submission format.
The CSVs need to include three additional fields:
- Affiliation
- Paradigm: Classical in your case
- Time to Solution

Please update the submission files accordingly. 
There is a [script](https://github.com/ZIB-AOPT/QOBLIB/blob/main/misc/migrate_submissions.ipynb) to automate the conversion, but I think in your case it's easiest if you update your scripts to add the additional information. 

Moreover, for stochastic algorithms, we require at least 5 runs per instance. 
The updated solution checker is available [here](https://github.com/ZIB-AOPT/QOBLIB/blob/main/misc/check_submission.py).
I'm sorry for the inconvenience. 

Best, Max

liuwb17:
Hi @mlschicker , 
I have resubmiited the results, please refer to "LABS submission revision 2". 
Best, Liu

timbode:
@liuwb17 Nice results, I will have a look!

PR #9, opened 2026-06-30
Add Qunova MIS submissions for c-fat200-1 and es60fst01-04

qunova-jr:
## Summary

Maximum Independent Set (MIS) submissions from Qunova Computing, Inc. for the
`07-independentset` benchmark, covering 5 instances. Solutions were
produced using IBM quantum hardware (ibm_marrakesh / ibm_miami).

Each instance now includes 5 solution runs (under `solutions/`), with the
summary CSVs and READMEs updated accordingly.

## Results (over 5 runs)

| Instance    | Optimum | Best set size | Best approx. ratio | Mean set size | Mean approx. ratio |
|-------------|---------|---------------|--------------------|---------------|--------------------|
| c-fat200-1  | 18      | 17            | 0.9444             | 16.0          | 0.8889             |
| es60fst01   | 60      | 60 (opt)      | 1.0000             | 59.0          | 0.9833             |
| es60fst02   | 88      | 86            | 0.9773             | 84.2          | 0.9568             |
| es60fst03   | 55      | 55 (opt)      | 1.0000             | 55.0          | 1.0000             |
| es60fst04   | 78      | 73            | 0.9359             | 70.8          | 0.9077             |


mlschicker:
Hi @qunova-jr ,

Thank you for your submission. 
As your algorithm is non-deterministic, would it be possible to provide at least 5 runs?

Best, Max

qunova-jr:
Hi @mlschicker ,

Thanks for the reply.
Yes, I will provide the additional results once completed.

qunova-jr:
Hi @mlschicker,

As you suggested, I have uploaded more results on all 5 instances and updated the above table accordingly.

Regards,
JR Lee

mlschicker:
Hi @qunova-jr ,

thank you for the update. 
All automated test pass. 

I have aligned the readmes in the instance directories with all the other submissions. 
(There is a command line argument to generate them automatically when checking the submission `--generate-readme`)

Can you describe your method in more detail please. 
I have a few specific questions here: 

> (1) The Independent Set problem is formulated as a binary optimization problem (QUBO);

I assume you use the penalty formulation with P = 2 that is also given in the models directory?
Please be precise here as there are many ways to formulate the same problem. 

> (2) Initial solution is computed calssically from the matrix’s structure as the initial “current best”

What method do you use here exactly?
What solution values do you get?
Would it be possible to provide and objective time series for this submission?
(There is a small typ: `calssically` -> `classically`) 

> (3) A parameterized quantum circuit is sampled, and each bitstring is decoded as a variable index (O(log n) qubits); (4) A candidate re-assignment for the sampled variables are computed classically, and accept if the global objective is improved; (5) Repeat step 3-4 while updating circuit parameters with classical optimizer until convergence or for predefined number of iterations;

Please be more specific on the method. 
If there is a paper where this is described in detail, you can provide a link. 

You can also add a README in the submissions main directory describing your method in detail. 

Best, 
Max

qunova-jr:
Hi Max,

Thanks for the review and the README cleanup — I'll use --generate-readme going forward.

(1) Yes — the penalty formulation with P = 2 from the models directory, used unmodified.

(2)–(5) On the method itself I have to ask for some patience. It's still pre-publication and our IP filings aren't complete, so at this stage I can't give an implementation-level description — that applies to the classical initialization as well as to the sampling and update steps, since the initialization is part of the method rather than a generic warm start. There's no paper to link yet either, but I'll come back to this PR with a reference as soon as one exists.

On the objective time series: attached is the convergence behaviour on the five MIS instances, approximation ratio against number of QPU calls, averaged over five runs per instance, with the band showing run-to-run min–max. Index 0 on the x-axis is the classical initial solution, so the plot also shows how much of the gap is closed by the quantum-classical loop.
image

I realize that leaves the entry less documented than the others. Does that fall below the bar for inclusion — is a submission possible at this level of detail, or does QOBLIB require the full method description up front? If there's a middle ground that would work, I'm glad to hear it.

Best,
JR

mlschicker:
Hi @qunova-jr , 

thank you for the update and clarification. 

Concerning the method description:
It is okay if the exact method is described later. 
You can add it via a PR whenever the work is published. 
This PR can be merged independently.

As for the objective time series:
Thank you for the nice plot!
Would it be possible for you to provide the raw data in our unified format?
An example can be seen in [this submission](https://github.com/ZIB-AOPT/QOBLIB/tree/main/07-independentset/submissions/20250715_Abs2_Schicker)

Moreover, make sure to fetch the newest changes from the main repo.

Best, 
Max

qunova-jr:
Hi Max,

Thanks for the review and the pointers!

I've added the raw objective time series in your unified format. A quick check
on whether I got the formatting right:

- I provide one gzipped JSON per run
  (`<instance>_objective_time_series_<n>.json.gz`, 0-indexed to match our
  per-run solution files), each containing `[[ {"Time", "Incumbent", "step"}, … ]]`.
- `Incumbent` is the negated size of the best feasible independent set found so
  far, and `step` is the QPU-call index (each sampler execution; step 0 is the
  classical initial solution).

One thing I'd like to flag: our runs only logged per-run total CPU/QPU times,
not a wall-clock timestamp at each individual step. So for now `Time` simply
mirrors the QPU-call index (i.e. `Time == step`). Please let me know if that's
acceptable, or whether you'd prefer us to instead distribute each run's total
runtime across its steps.

Could you also confirm the overall layout looks right? I used one file per run
to line up with our solution files — happy to consolidate into a single
per-instance file (outer list = runs) if that better matches the convention.

I've also rebased onto the latest `main`, and fixed the `calssically` →
`classically` typo you pointed out earlier — thanks for catching that!

Best,
Jeung Rac

mlschicker:
Hi @qunova-jr ,

thank you for the update. 

The submission checker flagged a few things:
- the objective time series should be given in a single json file for all runs. 
- the headers of the submission template do not match - this was changed recently

As for the time steps in the objective time series:
Please provide the wall-clock time stamps.
If they were equidistant between QPU calls, you can apply your suggestion. 

The submission checker will automatically update when you push changes to this PR. 

You can also run it locally via
```
uv run --project misc qoblib-check-submission <problem>/submissions/<name>
```

Please contact me, if you have any questions.

Best, 
Max

qunova-jr:
Hi Max,

Thanks for the feedback! I've updated the objective time series to your unified
format and fixed up the README/summary files accordingly.

Please let me know if anything else needs adjusting.

Best,
Jeung Rac

mlschicker:
Hi @qunova-jr ,

thank you for updating the submission. 
All automated tests pass now. 

I want to flag two remaining points:

1. Where did you get the optimal bounds?
Are they produced by your method, or dis you take the provided solution values in the repository?

2. You always state that `#Successful Runs = # Runs` but this is onl true if your success threshold is big enough to compensate for the non-optimal solutions. 
The Success Threshold is not provided, so I cannot check the validity of your claim.

Please add the Success Threshold to your data and report the number of successful runs accordingly. 

Best, 
Max

qunova-jr:
Hi Max,

Thanks for the review.

- Optimal bounds: these were taken from the provided solution values in the
  repository, not produced by our method — so I've set "Optimality Bound" to N/A.
- Success Threshold: I've added it (ε = 0) and corrected "# Successful Runs" to
  the number of runs that reached the optimum (0, 1, 0, 5, 0 across the five
  instances), with a note in Remarks.

Please let me know if anything else needs adjusting.

Best,
Jeung Rac

PR #13, opened 2026-07-08
Submission: Q-Bridge SwapSA — 07-independentset (4) and 01-marketsplit (4)

liveplex-cpu:
Dear QOBLIB maintainers,

following up on our email exchange with Max, this PR adds our benchmark submissions for two problem classes.

## Method
Constraint-preserving swap simulated annealing (quantum-inspired, classical hardware) — GPU-parallel SA with 512 chains, reheat rounds, and greedy-descent polish. Market Split additionally uses first-order persistency (roof-duality style) preprocessing. Full method summary in each submission README.

## Results (5 seeded runs per instance, seeds 42–46, 60 s/run, unfiltered statistics)
**07-independentset** — proven optimum matched in 5/5 runs on all 4 instances:
chesapeake (17), aves-sparrow-social (13), C125-9 (34), brock200-2 (12).

**01-marketsplit** — exact solutions (all Ax = b satisfied) in 5/5 runs on all 4 instances:
ms_03_050_002, ms_03_050_005, ms_03_100_001, ms_03_100_012 (~4.4 s/run).

## Validation
`misc/check_submission.py` passes 4/4 instances on both classes; per-instance READMEs generated with `--generate-readme`.

## Code availability
The engine is proprietary (commercial product); the algorithm is summarized in the READMEs and summary CSVs, and we are happy to provide additional detail or a review under NDA on request.

Thank you for the guidance — happy to adjust anything to match repository conventions.

Q-Bridge Team (South Korea) — liveplex@gmail.com


mlschicker:
Hi @liveplex-cpu ,

All automatic tests pass. 

Can you elaborate on what you mean by _Constraint preserving_ Simulated annealing?
I see that you put the constraints into the objective and solved the QUBO using SA. 
Moreover, you use post-processing that repairs the solution to be feasible, i.e., the constraints were not preserver during solve. 

Best, 
Max

liveplex-cpu:
Hi Max,

thank you for the careful read — you are right, and the wording deserved a fix.

*Constraint-preserving swap SA* is the name of our engine's mode for **assignment-structured problems** (problems with built-in exactly-one groups, e.g. vehicle-to-route assignment), where every move is a swap inside a group, so the search never leaves the constraint manifold. That mode is not applicable to MIS or Market Split and **was not used here**: for these two classes the engine ran in its standard mode — flip-move SA on the penalty QUBO, with (for MIS only) a deterministic repair step as post-processing, exactly as you describe. For Market Split no repair was involved — the reported solutions came out of the sampler with zero violation.

I have pushed a commit that makes the submission match what actually ran:
- renamed the submission directories from `20260709_SwapSA_QBridge` to `20260709_SA_QBridge`,
- scrubbed "constraint-preserving swap" from the READMEs, summary CSVs and solution headers (now: "GPU-parallel SA, flip moves on penalty QUBO"),
- added an explicit note in the MIS README that constraints are handled via the penalty term during the anneal and are not enforced move-by-move, with repair as post-processing.

`misc/check_submission.py` still passes 4/4 on both classes.

Thanks for catching this — please let me know if anything else needs adjusting.

Best,
Q-Bridge Team

mlschicker:
Hi @liveplex-cpu ,

thank you for the update!
All automatic tests pass. 

One final question:
Did you set the optimal objective value as a cutoff?
I am asking because your TTS and Runtime are equal in all runs. 
(Cutoff for marketsplit would be 0 for feasibility and the optimal known values for the independent set instances.)

If this is the case, please indicate this in the workflow. 

Thank you!

Best, 
Max

liveplex-cpu:
Hi Max,

good catch — no, we did **not** use the optimal value as a cutoff. The solver has no knowledge of optimal/feasible target values: it runs a fixed annealing schedule (512 chains, sweeps × reheat rounds) and exits early only when the wall-clock budget is reached. The varying runtimes across instances (e.g. 0.2 s on chesapeake vs ~21 s on C125-9) reflect the schedule size, not a target hit.

The reason TTS equalled Total Runtime in our CSVs is more mundane: we did not instrument time-to-solution separately and (incorrectly) filled both fields with total wall time. I have pushed a commit that fixes this honestly:
- **Time to Solution** is now reported as **N/A** in all 8 summary CSVs, and
- the **Workflow** column now states explicitly: "No optimality cutoff: the solver has no knowledge of optimal values and runs a fixed annealing schedule (early exit only on time budget). Time-to-solution was not separately instrumented (reported N/A)."

`misc/check_submission.py` still passes 4/4 on both classes. If you'd prefer we instrument first-hit times and re-run instead of N/A, happy to do that too.

Best,
Q-Bridge Team

mlschicker:
Hi @liveplex-cpu ,

okay, wonderful!
It would be interesting to see the TTS if possible. 

If you decide to rerun the experiments, it would be great if you could not only track the TTS but also the objective trajectory. 
The objective trajectory should be provided as a JSON. 
The format can be seen, e.g., in [this submission](https://github.com/ZIB-AOPT/QOBLIB/tree/main/02-labs/submissions/20250221_Abs2_Schicker/labs009).
I still have to specify it in the main guidelines. 
For now, the submission checker will check it for validity. 

If you decide not to rerun, no problem. 
Just leave a comment and I will merge the PR as is. 

Best, 
Max

liveplex-cpu:
Hi Max,

we decided to re-run — trajectories were worth having. Just pushed:

- **Objective trajectories**: `<instance>_objective_time_series.json` added for all 8 instances (5 runs each, `{Time, Incumbent}` points recorded at every incumbent improvement, per the format in the Abs2 labs009 submission). Incumbent semantics are documented in each README (penalty-QUBO objective; for MIS, at feasible incumbents it equals minus the set size).
- **Measured TTS**: the "Time to Solution" column is now computed post-hoc from the trajectories as the first time the final incumbent was reached (mean over 5 runs). It now clearly differs from total runtime — e.g. C125-9: TTS 7.1 s vs total 21.2 s, Market Split instances: TTS ~1.9 s vs total ~4.3 s — confirming the schedule always runs to completion with no cutoff.
- **Results unchanged** across the re-run: MIS 4/4 proven optima in 5/5 runs, Market Split 4/4 exact solutions in 5/5 runs.

`misc/check_submission.py` passes 4/4 on both classes including the time-series validation.

From our side the submission is complete — thank you for the thorough and helpful review!

Best,
Q-Bridge Team

mlschicker:
Hi @liveplex-cpu ,

perfect, thank you!

Everything looks good. 

Final question: 
You added an additional entry to the JSON at the final time step with the current incumbent, correct?

Best, Max

liveplex-cpu:
Hi Max,

correct — every run's trajectory ends with an explicit final entry recorded at termination time carrying the then-current incumbent, even when the value hasn't improved since an earlier point. For example, C125-9 run 0 ends with `{"Time": 7.076, "Incumbent": -34}` (last improvement) followed by `{"Time": 21.2339, "Incumbent": -34}` (termination). The recording points are: initialization, every incumbent improvement, and termination (after the final greedy-descent polish).

Best,
Q-Bridge Team

Issue #2, opened 2026-02-24
Steiner tree packing instances may be trivial for heuristic solvers


RoutingEnthusiast:
Unless I am mistaken the Steiner tree instances are easily solved by simple ripup and reroute approaches (non-optimally of course). The first variant I tried could generate valid solutions for all instances without any tuning, according to the provided checker. Looking at a few instances it seems that there is too much space or too few terminals. 

A solution file for instance stp_s100_l4_t6_h2_rs37235 is [s100_l4_t6_h2_rs37235.bst.sol.txt](https://github.com/user-attachments/files/25532811/s100_l4_t6_h2_rs37235.bst.sol.txt). It has length 2023. Note that this is on a 100x100x4 grid so it is almost empty. 

Should you be able to provide denser and more difficult instances I would be very happy, especially if they were closer to modern VLSI-routing instances.

RoutingEnthusiast:
Just a note on what's out there: For my experiments I ended up using the benchmarks from ReSPack (https://doi.org/10.17605/OSF.IO/BJ9R3). These instances were a bit more difficult and some were much larger but they were ultimately also easy to solve non-optimally. That's for the instances without margins. I did not implement margins in my heuristic solver. Other nice instances can be found in the ISPD detailed routing contests (but if I remember correctly they had design rules that I did not want to implement as I was just interested in raw Steiner tree packing).

PR #18, opened 2026-07-13
STP generation

gnannicini:
Refactored generation script. Added a generation script that tries to make things harder for simple decomposition heuristics.

### DIFF
--- a/04-steiner/misc/README.md
+++ b/04-steiner/misc/README.md
@@ -12,6 +12,51 @@ Generates feasible Steiner tree packing instances on a multi-layer grid with con
 - **ZIMPL**: Modeling language for mathematical programming ([SCIP Optimization Suite](https://www.scipopt.org/#download))
 - **CPLEX**: IBM optimization solver ([IBM CPLEX](https://www.ibm.com/products/ilog-cplex-optimization-studio))
 - **Python packages**: `numpy` (install via `pip install numpy`)
+- **Models**: ZIMPL file stp_gen_blocked.zpl
+
+**Usage:**
+```bash
+python3 generate_feasible_instance.py <size> <num_layers> <max_terminals> [options]
+```
+
+**Arguments:**
+
+| Argument            | Description                                                          | Default  |
+| ------------------- | -------------------------------------------------------------------- | -------- |
+| `size`              | Grid size (for size=n, creates n×n grid)                             | Required |
+| `num_layers`        | Number of stacked grid layers                                        | Required |
+| `max_terminals`     | Max terminals per net (reduced iteratively until feasible)           | Required |
+| `--num_holes`       | Number of random holes in the grid                                   | `0`      |
+| `--randseed`        | Random seed for reproducibility                                      | `12345`  |
+| `--zimpl_path`      | Path to ZIMPL executable                                             | `zimpl`  |
+| `--cplex_path`      | Path to CPLEX executable                                             | `cplex`  |
+| `--force_diff_side` | Force second terminal of each net to be on different side from first | `False`  |
+
+**Example:**
+```bash
+python3 generate_feasible_instance.py 10 3 5 \
+    --num_holes 2 \
+    --randseed 123 \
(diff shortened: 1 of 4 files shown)


mlschicker:
Hi @gnannicini ,

thank you for submitting the new STPP generation script and also updating the old one!

Thi seems to be a good way to resolve Issue #2 .
It would be good if we could verify the new instance generation script extensively with the code used in the Issue. 
On this note, I would ask @RoutingEnthusiast to also take a look at this and give us their thoughts. 

@gnannicini : Which instances would you suggest we add to the repository?

Looking forward to this discussion!

Best, 
Max

RoutingEnthusiast:
Thanks! I would be happy to run this with my simple heuristic solver but I would need the instances (not just the code for generating them), as I don't have access to cplex here.

gnannicini:
Regarding what instances to try, I only tried instances up to 30x30. Not sure at what size we expect @RoutingEnthusiast 's heuristic to stop being effective. In the couple of instances of size 30x30 that I tried, Gurobi could find feasible solutions even better than the "seed" partial solution given in the files, but that's not the point of the modified instance generator: the point of the modified generator is that it is *less likely* (not guaranteed) that a simple decomposition with optimal steiner strees between individual terminals would give you the optimum. As with all tests, I think starting from relatively small sizes (20x20 or 30x30) would be reasonable, just to see if there is a noticeable difference compared to the old generator.

A quick note on the solver: the script uses CPLEX, it should be very easy to change this to use Gurobi instead. The models are written with ZIMPL, so the only modifications needed are to change the command line instruction to run Gurobi instead of Cplex, and to change the function that reads a CPLEX solution file to a function that reads a Gurobi file. A good AI coding tool can probably do that flawlessly. (But I didn't try--this code is written entirely by hand.)

RoutingEnthusiast:
Sorry I don't have Gurobi here either. If there are reasons not to post instances I can try with an inferior solver (probably this weekend).

I would guess that are very small instances (10x10 and even smaller) that are solvable but that are very difficult for heuristic solvers... So I'm really interested in seeing your instances!

RoutingEnthusiast:
As mentioned I didn't try with gurobi / cplex and maybe I was using it wrong but I got instances like this (instance + solution by my simple heuristic solver):

layer 1:
distinct-layer-1

layer 2:
distinct-layer-2

These instances are too easy to be useful to me. I would be interested in non-trivial instances like the one below and in more difficult instances (instances with only 2-terminal nets as in this example and instances without this restriction). I'd be even more interested in non-trivial non-switchbox instances.

layer 1:
distinct-layer-1

layer 2:
distinct-layer-2

For this instance my simple heuristic solver fails. To generate this instance I started with an empty instance and iteratively tried adding a 'random' 2-terminal net, locating the two terminals on different sides. I accepted a new net if a (better) solver was still able to solve the instance with the additional net.

I would be curious if cplex / gurobi can find a solution to this tiny instance in a reasonable amount of time (my guess is no but I'm unsure). In case someone wants to play with this instance I attached the instance files: 
[less_trivial_instance.tar.gz](https://github.com/user-attachments/files/30373841/less_trivial_instance.tar.gz)

mlschicker:
@gnannicini 
Can you take a look at this please?

@RoutingEnthusiast 
It seems that these instances only have two terminals per network.
We want multiple terminals per network.

RoutingEnthusiast:
```
@RoutingEnthusiast
It seems that these instances only have two terminals per network.
We want multiple terminals per network.
```

@mlschicker 
If I remember correctly the script from the pull request has a parameter for controlling that but instances with more terminals weren't more interesting.

If you were talking about the instances that I generated with the procedure I describe above: And you are correct that these don't have 3+ terminal nets. It should be easy to generate similar instances with more terminals per net (but instances with multiple terminals at the same side may behave quite differently). 

If you are interested in including them in your benchmark, I can try to generate instances of various sizes that I would consider more interesting, but I would probably not be interested in sharing the code for generating them.

PR #19, opened 2026-07-22
Automate submission validation, best-known-value tracking, and site builds

mlschicker:
## Summary

This PR reworks the contribution pipeline so that submissions are validated,
credited, and published automatically. Contributors no longer touch any
`solutions/` directory by hand — CI checks each PR, comments the result, and
recomputes best-known values (BKV) on merge. It also reorganizes `misc/` into
clear areas and moves the Python tooling onto `uv` with a committed lockfile.

## What's included

**Contributor-facing**
- Issue templates (bug report, feature request, submission support) and split PR
  templates for instance additions and submission results.
- Expanded `CONTRIBUTING.md`: an explicit, declaration-based solution policy
  (a file must always be valid; feasibility/optimality are only enforced when the
  CSV claims them) and a note that BKV tables update automatically on merge.

**CI / workflows** (`.github/workflows/`)
- `validate-submission.yml` — runs the authoritative checker on changed
  submissions and posts a sticky PR comment; a red run blocks merge.
- `validate-submission-comment.yml` — updates that comment in place instead of
  posting duplicates.
- `update-bkv.yml` — on merge, recomputes each problem's best-known value from
  reference solutions + feasible submissions, copies improving solutions into
  `NN-problem/solutions/`, and regenerates the attributed BKV tables.
- `tests.yml` and `pages.yml` split so unit tests and site deployment run
  independently.

**Tooling reorg** (`misc/`)
- `misc/ci/` — pipeline scripts (`check_submission.py`, `check_bkv_updates.py`,
  `update_bkv.py`, `build_site.py`, `render_comment.py`) plus the shared
  `site_builder/` package, each exposed as a `qoblib-*` console script.
- `misc/tools/` — standalone utilities; `misc/notebooks/` — migration notebook.
- `CHECKER_CONTRACT.md` documents the exit-code contract shared by the Rust
  per-problem checkers and `check_submission.py`.
- `local-ci.sh` reproduces every workflow locally without Docker.
- Switched to `uv`: `pyproject.toml` + `uv.lock` (committed), scripts depend only
  on the Python standard library.

**Housekeeping**
- Regenerated `solutions/README.md` BKV tables across problem classes.
- Minor fixes to per-problem Rust checkers and `gen_archive.sh` scripts.

## Testing it locally

```bash
misc/ci/local-ci.sh            # everything
misc/ci/local-ci.sh validate   # check_submission + check_bkv against origin/main
misc/ci/local-ci.sh bkv        # update_bkv --check (non-destructive)

### CODE REVIEW

**Comment by mlschicker** [comment]
Thanks Copilot — went through all the findings. Summary of what changed:

Absolute runner paths in issue/PR templates (submission_support, bug_report, feature_request, instance_addition) — fixed, all now use repo-relative `CONTRIBUTING.md` (6 occurrences).

config.yml contact links — repointed from mlschicker/QOBLIB to ZIB-AOPT/QOBLIB.

Checker exit-code contract (marketsplit, birkhoff, topology, stableset) — agreed, the blanket panic-hook→10 was mislabeling infrastructure errors as INVALID_FILE. USAGE cases now exit 2 per misc/ci/CHECKER_CONTRACT.md:

marketsplit: bad args, unreadable instance file, unreadable solution file → 2
birkhoff: instance read + instance-JSON parse + solution read → 2 (malformed solution JSON still → 10)
topology: node-count/degree/diameter CLI parse errors → 2
stableset: unreadable solution file → 2, plus the shared Graph::read_from_file File::open failure → 2 (this is the "unreadable instance file looks like a bad solution" case; applied to topology's copy of the fn too)
Malformed solution content still funnels to 10 via the panic hook in all four. Verified the checkers build and that no-args / missing-instance / bad-numeric-arg all return 2.

PR-number from untrusted artifact (validate-submission-comment.yml) — real issue, fixed with a different approach than suggested. context.payload.workflow_run.pull_requests[0] is empty for fork PRs, which is the case this workflow exists to handle, so it would silently stop commenting on real submissions. Instead I keep the artifact's PR number but verify it against the trusted workflow_run.head_sha:

```
let pr;
try {
  ({ data: pr } = await github.rest.pulls.get({ owner, repo, pull_number: prNumber }));
} catch (e) {
  core.info(`PR #${prNumber} from artifact not found (${e.status}). Exiting.`);
  return;
}
const runHeadSha = context.payload.workflow_run.head_sha;
if (pr.head.sha !== runHeadSha) {
  core.warning(`Artifact PR #${prNumber} head ${pr.head.sha} != run head ${runHeadSha}; refusing to comment.`);
  return;
}
```
A malicious PR that stamps a victim's number into pr-number.txt is rejected, because that PR's head SHA won't match the attacker's run. Fork-safe and spoof-safe.

Also fixed while in here (not flagged): bug_report.yml was invalid YAML — description: Example: 01-marketsplit and placeholder: Example: msc_100_5 have an unquoted :  that would prevent the form from loading. Both quoted.

stefan-woerner:
👍

PR #22, opened 2026-07-24
Add Greedy LS submission

mlschicker:
# Kipu Quantum submission

@NarendraHegade5 : please add the necessary description.


bernalde:
It is close to merging, likely one small metadata revision away from approval.
Blocking: Missing provenance. All 34 submission CSVs have an empty Reference field, although QOBLIB requires a paper, repository, or detailed documentation covering hyperparameters, software versions, and implementation details. One appropriate reference can be repeated across the submissions.
Blocking: Ambiguous frb100-40 success count. The summary CSV reports 10 successful runs and Success Threshold = N/A. The trace contains final best values 94, 95, 94, 95, 94, 94, 95, 95, 94, 95; only 5 runs reach 95. They should either report 5 successes with threshold 0, or provide the numerical threshold under which 94 counts as successful—at least 1/95 ≈ 0.01053.
Nonblocking: PR description. The body is still a placeholder asking for the necessary description and does not follow the submission template. It should summarize the algorithm, scope, reference, and validation performed.

PR #24, opened 2026-07-30
upload instances for quantum benchmarks

### LINKED ISSUE #23: [Feature]: New instances requested, existing portfolio optimization instances are intractable in current quantum hardware
### Scope

Instances

### Related problem class (if any)

06-Portfolio

### Motivation / problem

At Global Data Quantum, we have been exploring quantum and hybrid approaches for the multi-period Portfolio Optimization problem included in QOBLIB.

While implementing the corresponding QUBO formulation (QOBLIB paper (<https://arxiv.org/abs/2504.03832>), Current UQO/QUBO implementation (<https://github.com/ZIB-AOPT/QOBLIB/tree/main/06-portfolio/models/unconstrained_quadratic_optimization>)), we observed that the existing benchmark instances start at a problem size that is substantially larger than the raw qubit count of currently available quantum processors. We therefore propose adding a complementary set of smaller, hardware-scaled benchmark profiles containing 20, 56, 120, and 148 binary variables.

**Current problem-size gap**

The portfolio formulation uses a unary encoding for long and short asset positions. For $n$ assets, $m$ time periods, and at most $k$ units of each asset, the asset-position register contains $2kn$ binary variables per time period.

The capital and portfolio cardinality constraints introduce two additional binary registers. When these registers are sized for the parameters of each instance, the resulting number of binary variables is:

$$N_{bin} = m\left( 2kn + \left\lceil \log_{2}(B + 1) \right\rceil + \left\lceil \log_{2}(C + 1) \right\rceil \right)$$

where:

- $B$ is the maximum number of simultaneously selected asset positions;
- $C$ is the number of normalized capital units.

Under a direct QUBO-to-Ising encoding, each binary variable corresponds to one logical qubit before transpilation, routing, or the introduction of any auxiliary qubits.

**Existing QOBLIB instances**

| $\mathbf{n}$ | $\mathbf{m}$ | $\mathbf{k}$ | $\mathbf{B}$ | $\mathbf{C}$ | Variables reported in QOBLIB | Instance-specific binary variables |
|---|---|---|---|---|---|---|
| 10 | 10 | 3 | 4 | 10 | 710 | 670 |
| 10 | 15 | 3 | 4 | 10 | 1,065 | 1,005 |
| 50 | 10 | 3 | 20 | 10 | 3,110 | 3,090 |
| 50 | 15 | 3 | 20 | 10 | 4,665 | 4,635 |
| 200 | 10 | 3 | 50 | 10 | 12,110 | 12,100 |
| 200 | 15 | 3 | 50 | 10 | 18,165 | 18,150 |
| 400 | 10 | 3 | 100 | 10 | 24,110 | 24,110 |
| 400 | 15 | 3 | 100 | 10 | 36,165 | 36,165 |

The difference between the two final columns comes from the cardinality slack register. The current QOBLIB parameterization fixes this register to seven bits, based on the largest value ($B = 100$), whereas the instance-specific column uses only the number of bits required by each value of $B$.

Consequently, the smallest existing instance contains:

- 710 binary variables as currently distributed, or
- 670 binary variables with instance-specific slack-register sizing.

Both counts exceed the raw capacity of current 120-qubit Nighthawk and 156-qubit Heron processors.

### Proposed change

**Proposed size tier**

We propose adding the following complementary benchmark sizes with the following parameter combinations.

The following parameter combinations produce the requested sizes:

| $\mathbf{n}$ | $\mathbf{m}$ | $\mathbf{k}$ | $\mathbf{B}$ | $\mathbf{C}$ | $\mathbf{N_{bin}}$ | Intended role |
|---|---|---|---|---|---|---|
| 3 | 2 | 1 | 3 | 3 | 20 | Local simulation, debugging, and QPU smoke tests |
| 4 | 4 | 1 | 4 | 7 | 56 | Intermediate simulation and reduced QPU experiments |
| 4 | 4 | 3 | 4 | 7 | 120 | Raw-qubit-count alignment with Nighthawk processors |
| 5 | 4 | 3 | 4 | 10 | 148 | Raw-qubit-count alignment with Heron processors |

This would provide a gradual progression between a classically inspectable problem and the current hardware limit. Note that the 56- and 120-variable profiles use the same number of assets and time periods but different values of $k$.

We propose retaining the financial assumptions used in the original benchmark wherever they are independent of the instance size:

| **Parameter** | **Proposed value** |
|---|---|
| Risk-free interest rate $\nu$ | 0.01% |
| Transaction-cost rate $\delta$ | 0.1% |
| Short-position loan rate $\rho$ | 0.0025% |
| Total available cash | USD 1,000,000 |

**Proposed initial values of $\lambda$ and $P$**

| **Binary variables** | **Risk coefficient ($\lambda$)** | **Penalty multiplier $P$** |
|---|---|---|
| 20 | $1 \times 10^{-5}$ | 5.0 |
| 56 | $5 \times 10^{-5}$ | 1.0 |
| 120 | $2 \times 10^{-5}$ | 1.0 |
| 148 | $4 \times 10^{-5}$ | 0.5 |

These values were obtained through an initial coefficient-scale analysis:

- $\lambda$ was adjusted so that the risk contribution remains comparable in scale to the return contribution;
- $P$ was selected so that the capital and cardinality constraints receive coefficients approximately one order of magnitude larger than the principal non-penalty terms.

**Reproducible generation procedure**

The existing QOBLIB generator can be reused for the generation of the instances, following the guidelines of the original instances (see <https://github.com/ZIB-AOPT/QOBLIB/tree/main/06-portfolio/misc/instance_generation>).

### Alternatives considered

Even though larger quantum processors could be considered, currently available hardware does not provide the qubit counts required by the existing instances. Reducing the qubit count through reformulation may also be possible, but incremental changes to the encoding may not be sufficient; achieving hardware-compatible sizes could require a substantially different, problem-specific modeling approach and a broader reconsideration of the portfolio optimization formulation. Such a redesign is outside the scope of this issue.

### Expected impact

These smaller instances would enable contributors to run the portfolio benchmarks on real quantum hardware and compare the resulting solutions and performance against classical baselines. They would also provide reproducible, hardware-relevant test cases for future quantum and hybrid submissions.

### Confirmation (do not remove)

- [x] I checked existing issues for duplicates.
- [x] I included concrete motivation and scope.

DanielGarcia-GDQ:
**Summary**

This pull request adds a new set of smaller benchmark instances for the Portfolio Optimization problem (06-portfolio) to complement the existing QOBLIB benchmark suite.

**Changes**
Added four new benchmark instances with:

- 20 binary variables
- 56 binary variables
- 120 binary variables
- 148 binary variables

Generated the instances following the existing QOBLIB instance generation procedure and benchmark conventions.
Fixes #23


DanielGarcia-GDQ:
cc: @DanelArias-Dreyton257 @manuelmartin3 @QNod

mlschicker:
Hi @DanielGarcia-GDQ , 

Is it correct that these instances were generated using the provided instance generator?
Please indicate the `numpy` version you used. 

Blocking: Please add the license header to all of the new instances. 

Best, 
Max

DanielGarcia-GDQ:
Hello @mlschicker,

I have pushed a new commit that updates the instance files by adding the license header.

Additionally, I have verified the NumPy version currently being used. The version is **1.26.4**.

PR #25, opened 2026-07-30
ParityQC QOBLIB's Submission

maxC-PQC:
In this PR, we incorporate performance benchmarks for ParityQC's [Parity Twine Optimizer](https://parityqc.com/products/parity-twine-optimizer) into QOBLIB.

Evaluation was conducted for the `market-split` and `independent-set` problems.

## Validation
misc/check_submission.py passes on all instances.


mlschicker:
Hi @maxC-PQC ,

All automated tests pass. 

For all instances, you report `Success Threshold = N/A` and `# Successful Runs = 0`.
Please change this. 
For `marketsplit`, change the threshold to `0` and set `# Successful Runs = # Feasible Runs`.
(That is, if you model the problem as a QUBO, which I assume)
For `independent-set` you can choose. I suggest, setting the threshold to `0` and set `# Successful Runs` to the number of runs that reached the known optimum. 

Could you go into a bit more detail in the Workflow, e.g., which model you used. 

Moreover, would it be possible for you to provide an objective time series?
The format is a simple JSON file like

```
[
  [
    {"Time": 0.001,  "Incumbent": 120.0},
    {"Time": 0.218,  "Incumbent": 5.0},
    {"Time": 0.435,  "Incumbent": 0.0}
  ],
  [
    {"Time": 0.002,  "Incumbent": 80.0},
    {"Time": 0.651,  "Incumbent": 0.0}
  ]
]
```

Thank you!

Best, 
Max

maxC-PQC:
> Hi @maxC-PQC ,
> 
> thank you for your submission. All automated tests pass.
> 
> For all instances, you report `Success Threshold = N/A` and `# Successful Runs = 0`. Please change this. For `marketsplit`, change the threshold to `0` and set `# Successful Runs = # Feasible Runs`. (That is, if you model the problem as a QUBO, which I assume) For `independent-set` you can choose. I suggest, setting the threshold to `0` and set `# Successful Runs` to the number of runs that reached the known optimum.
> 
> Could you go into a bit more detail in the Workflow, e.g., which model you used.
> 
> Moreover, would it be possible for you to provide an objective time series? The format is a simple JSON file like
> 
> ```
> [
>   [
>     {"Time": 0.001,  "Incumbent": 120.0},
>     {"Time": 0.218,  "Incumbent": 5.0},
>     {"Time": 0.435,  "Incumbent": 0.0}
>   ],
>   [
>     {"Time": 0.002,  "Incumbent": 80.0},
>     {"Time": 0.651,  "Incumbent": 0.0}
>   ]
> ]
> ```
> 
> Thank you!
> 
> Best, Max

These changes have been made.
We however did not track the time of the best incumbent.

mlschicker:
Hi @maxC-PQC , 

thank you for the update. 
Which QUBO formulation did you use?
The canonical one with a penalty of 2?

Moreover, did you set the optimal known objective value as a cutoff?
(I assume yes as TTS = Runtime)

Thank you!

Best, 
Max

maxC-PQC:
> Hi @maxC-PQC ,
> 
> thank you for the update. Which QUBO formulation did you use? The canonical one with a penalty of 2?
> 
> Moreover, did you set the optimal known objective value as a cutoff? (I assume yes as TTS = Runtime)
> 
> Thank you!
> 
> Best, Max


Hi @mlschicker, 
We used a formulation of the QUBO with penalty (n / d) * log(d) where n is the number of nodes and d the average degree of the graph. I added this in the submission.

We didn't set any cutoff, the optimizer runs to completion and then results are reported.

Kindly,
Max

mlschicker:
Hi @maxC-PQC , 

Is there a reason you chose this penalty over the canonical penalty of 2?

Moreover, please indicate the coefficient range in each of the summaries. 
As an additional detail, please report the penalty used (although this could also be extracted from the Coefficients Range).

Is the number of non-zero coefficients correct?
The canonical QUBO formulation has |V| + |E| non-zero coefficients. 
Your formulation seems to only have |E| non-zeros. 
Please verify this detail. 

The same points also apply to your marketsplit submission.
Could you provide the exact model you used for completeness please?

Thank you!

Best, 
Max

maxC-PQC:
> Hi @maxC-PQC ,
> 
> thank you for the clarification. Is there a reason you chose this penalty over the canonical penalty of 2?
> 
> Moreover, please indicate the coefficient range in each of the summaries. As an additional detail, please report the penalty used (although this could also be extracted from the Coefficients Range).
> 
> Is the number of non-zero coefficients correct? The canonical QUBO formulation has |V| + |E| non-zero coefficients. Your formulation seems to only have |E| non-zeros. Please verify this detail.
> 
> The same points also apply to your marketsplit submission. Could you provide the exact model you used for completeness please?
> 
> Thank you!
> 
> Best, Max

We followed the approach of https://arxiv.org/pdf/1007.1378.
The coefficient ranges are now included. 
There was an accounting error for the number of terms in ms and mis. 

For the market-split,  we used the canonical form of the qubo.
Thanks,

Max

mlschicker:
Hi @maxC-PQC ,

thank you for the clarification and for updating the coefficient ranges. 

The paper you provided only talks about random graphs. 
Moreover, it does not say anything directly about which penalty to choose. 
Can you go into more detail why you chose this penalty factor please?
I am curious. 

Thank you!

Best, 
Max

maxC-PQC:
Hi Max,

We had explored multiple values for the penalty >2 and did not unfortunately observe a dependence on the final performance, and so we kept this penalty.
We note that in the case of the problems submitted, the penalty is greater than 2 for all instances.

Kindly,

PR #27, opened 2026-08-02
[03-birkhoff] Add QAOA E-FCFW benchmark solutions

stephendiadamo:
### Submission Summary

This Pull Request adds benchmark solutions for **03-birkhoff** (Minimum Birkhoff Decomposition) using Partitioned QAOA + E-FCFW with edge-swap local refinement executed on **IBM Pittsburgh** via Qoro Cloud Services.

This algorithm and benchmarking work were developed in collaboration between **E.ON** and **Qoro Quantum**.

#### Submitted Instances:
- `qbench_16_sparse_id6`: Best Objective Value `5.707420e-03`
- `qbench_32_sparse_id1`: Best Objective Value `1.737801e-03`
- `qbench_64_sparse_id1`: Best Objective Value `7.840509e-04`

#### Verification:
- Automated submission validator: `Summary: 3/3 instances passed, 0 failed. Overall: OK`
- Best-Known-Value preview: verified new records for all 3 instances.


mlschicker:
Hi @stephendiadamo ,

There are a few blocking issued with the submission as it stands:
- Please use the names of the instances as the names of the directories in the submission. 
The official instance names are the IDs, i.e., `B16_16_1` would be the instance of size 16, sparse, id 1. 
I will update the submission checker to make name checks like this mandatory. 
You can take a look at [this submission](https://github.com/ZIB-AOPT/QOBLIB/tree/main/03-birkhoff/submissions/20260111_Gurobi_Schicker) for the general structure. 
- You solution files are not in the correct format yet. The submission checker will check them after you have updated the directory structure with the correct instance names. 
- There are stray summary files in the directory. 
Please use the structure you see in the submission I linked above. 

For `B16_16_6`, I reformatted and checked your solution. 
The first permutation matrix provided in the solution file reads
```
[10,11,9,7,15,0,13,3,0,2,0,1,14,6,12,4]
```
According to our format, this is not a permutation as it includes index 0 three times, and no 5 or 8. 
Did you use a different format to describe permutation matrices?

Please describe your method in more detail and provide a link to the paper. 

Best, Max

stephendiadamo:
Hi Max,

I've used the wrong format. The outputs from our routines are symmetric adjacency matrices rather than permutation matrices. 

I will need to rerun these experiments unfortunately. I will run a smaller instance with simulation first to make sure the formats are correct and then will rerun them on the hardware. 

We haven't yet published our paper on the method, but the main method is described in the original paper. The changes are mostly in the partitioning and recombination. I'll expand on the method details.

Thanks,
Stephen

mlschicker:
Hi @stephendiadamo , 

thank you for the quick response!

You can update this PR once you are done rerunning the experiments. 
With the updated directory structure, the checks should catch any remaining inconsistencies. 
Sorry for not catching this earlier - as mentioned above, I will update the script accordingly. 

You can put the details into a README in the submission directory. 
Moreover, make sure to follow the naming convention of the submission directories, i.e., `YYYYMMDD_<METHOD_SHORT>_<LAST_NAME>`.

Please don't hesitate to ask, if you have further questions. 

Best, 
Max

stephendiadamo:
I'm working on getting the submission correct, I'll post the final results shortly.

mlschicker:
Hi @stephendiadamo , 

thank you for updating the submission. 

```
| Algorithm Type | Quantum-Classical Hybrid |
| Paradigm | QAOA |
```
The Paradigm should be either `Classical`, `Quantum Simulator`, `Quantum Hardware`
The Algorithm Type should be either `Stochastic` or `Deterministic`. 

I assume that your algorithm is stochastic. 
In this case, we require at least 5 runs per instance, where you report the best runtime. 

You mention in the hardware specifications that you used `ibm_pittsburgh` but the QPU runtimes are zero. 
Did you use a simulator instead?

Best, 
Max

stephendiadamo:
I have the redone hardware execution and hopefully now all the formats are correct. 

One thing is, I did not run it 5x since it uses QPUs and can be quite expensive. I could run it 5x in simulation to compare. Perhaps I can have it in a second row in the same submission.

mlschicker:
Hi @stephendiadamo , 

quick question:
Which metric do you use to measure the approximation to the desired matrix?

Best, 
Max

stephendiadamo:
We use the normalized squared Frobenius norm. 

It's defined in equation 2 here: https://arxiv.org/pdf/2509.10657

mlschicker:
Thank you for the clarification. 

If you provide further runs, would it be possible to also provide the objective time series?

Best, 
Max

stephendiadamo:
Yes, it's no problem. I will add in this in after running the final 4 repetitions.

mlschicker:
Hi @stephendiadamo , 

please remove the `.DS_Store` artifacts. 
They probably come from extracting compressed directories. 

Moreover, I checked some of your results and am a bit confused. 
For each instance, I have computed the normalized squared Frobenius norm, i.e., `f(X) = (1 / n^2) * || D - X ||^2_F` for `X = 1_(n,n)/n` (the uniform double stochastic matrix).

```
Instanz              | Your Objective |   X = 1/n | Comment
qbench_16_sparse_id6 |   5.707420e-03 |  0.006997 | OK
qbench_32_sparse_id1 |   1.737801e-03 |  0.001435 | Worse
qbench_64_sparse_id1 |   7.840509e-04 |  0.000490 | Worse
```

Which metric is used in the code?

Best, 
Max

stephendiadamo:
Hi Max, you got those results from the first commit? I wrote a script that was taking data from a local cache. Those numbers are from a previous run. I was doing some initial testing and it picked up the wrong data by repeat runs where I used fewer iterations to test the pipeline. They've been removed in the following commits.

The objective value in the summary CSVs now correctly reflects `k` (the number of permutation matrices: 66, 164, and 280).

The objective function minimized internally by CPLEX during the algorithm iterations is indeed the normalized squared Frobenius norm. I'll post the convergence trend.

mlschicker:
I see, thank you for the clarification. 
What exactly is the success threshold here?
The checker claims that the decompositions are exact.

stephendiadamo:
Hi Max, I was previously using our algorithm's internal early-termination tolerance as the "Success Threshold", but now I see that field wasn't intended for solver tolerances. Since we run a fixed 15 iterations with exact completion, I've updated that field to N/A.

The submitted decompositions are 100% mathematically exact, which is now explained in the main README. Briefly, our pipeline runs 15 iterations of Partitioned QAOA sampling within E-FCFW, and any remaining residual matrix entries are cleared using the classical BKF algorithm via Hungarian matching (scipy.optimize.linear_sum_assignment). The E-FCFW routine with 15 iterations clears ~99% of the weight and then Hungarian algorithm finishes the task.

mlschicker:
HI @stephendiadamo , 

thank you for the update and clarification.

Please convert the objective time series to the correct format: 

```
[
  [
    {"Time": 0.001,  "Incumbent": 120.0},
    {"Time": 0.218,  "Incumbent": 5.0},
    {"Time": 0.435,  "Incumbent": 0.0}
  ],
  [
    {"Time": 0.002,  "Incumbent": 80.0},
    {"Time": 0.651,  "Incumbent": 0.0}
  ]
]
```

You can also add additional fields to `Time` and `Incumbent` if you have more interesting information. 
It would be great if you could provide the time series for all of the runs in the above format (one list entry per run).

Do the solutions differ in your runs?
If yes, please provide all solution files.

There is a stray `.DS_Store` file in the repository, please remove. 

What CPU do you run the classical part on?
Did you record CPU time?
What happens during the non-QPU time?

Best, 
Max

stephendiadamo:
Hi Max, the results are of the best of the 5 trials per iteration. 

The solutions do differ run to run but I didn't store it. I would have to rerun it to more solution files. I can run 1 or 2 more if it is necessary.

> What CPU do you run the classical part on? 

I run this on my Mac Mini.

> What happens during the non-QPU time?

The CPU part prepares the job using Divi and some helper functions, submits it to Qoro's service, and then waits for the QPU results to come back. This is the partitioning part, QUBO generation, an API call to characterize the QUBOs (0.1s per call),  conversion to QAOA circuits, and then submission. It then waits for the the results to come back. Then it performs the "stitching" parts via beam search, the post processing and then the CPLEX optimization routine.  I considered these negligible timings, < 5 minutes for the entire run outside of waiting for the QPU results on a personal computer. 

> Did you record CPU time?

Because there is nothing very significant happening on the CPU side, I didn't record the time. If needed I can take sample using simulation to time just the classical parts.

The total time was recorded, but last night while I was running it, the QPU went out for maintenance so it took longer than it could have.

Note: I don't have a value for the time, so I just put 0 everywhere.

mlschicker:
Hi @stephendiadamo , 

thank you for the clarifications. 

> I run this on my Mac Mini.

Which CPU Model does your Mac Mini have?
I am asking because performance differs a lot across models and I am not sure how much of your algorithm ran on CPU.

> I considered these negligible timings, < 5 minutes for the entire run outside of waiting for the QPU results on a personal computer.

`<5 minutes = 300 s` on CPU is not negligible. 
Especially when your QPU runtime is of the same order. 
Moreover, you report overall runtimes that far exceed the QPU times.
Did you include queuing times?
If so, please do not include them in the overall runtime.

> Briefly, our pipeline runs 15 iterations of Partitioned QAOA sampling within E-FCFW, and any remaining residual matrix entries are cleared using the classical BKF algorithm via Hungarian matching (scipy.optimize.linear_sum_assignment).

You describe that part of the method runs classically. 
So we should definitely report the time it ran on classical hardware.

> The solutions do differ run to run but I didn't store it. I would have to rerun it to more solution files. I can run 1 or 2 more if it is necessary.

Do the solutions also differ in the objective value?
It is not necessary to include all of the solutions, just interesting. 

What does the `Incumbent` field in the objective time series data reflect?
As the problem is `Minimum Birkhoff Decomposition`, the incumbent should be the number of matrices currently in the decomposition. 
The approximation should be a different field here. 

Thank you!

Best, 
Max

omearac:
@stephendiadamo here's an example of the CSV with the metadata for the submission regarding CPU details and etc: https://github.com/victor-ibm/QOBLIB/tree/E-FCFW_August_2026_Pennington_Mohseni/03-birkhoff/submissions/20260804_E-FCFW_Pennington-Mohseni/B10_10_1

omearac:
@stephendiadamo for the time series incumbent checkout victors Birkhoff+ for this problem showing thats the current number of items in the decomposition: https://github.com/victor-ibm/QOBLIB/blob/Birkhoff%2B_August_2026/03-birkhoff/submissions/20260803_BirkhoffPlus_Valls/B64_64_1/B64_64_1_objective_time_series.json

stephendiadamo:
@mlschicker 

I have a Mac Mini with a 2024 with M4 Pro processor, 48 GB memory. It's relatively performant. 

Yes, queue times were recorded in the total. I will deduct them. I'll run this in simulation and record the CPU time to get a good estimate... I would want to avoid running the whole thing with the QPU again to get the exact CPU time...

mlschicker:
Hi @stephendiadamo , 

okay, that sounds good!
Your suggestion with the simulator sounds good. 

Ping me once the results are updated. 

Thank you!

Best, 
Max

stephendiadamo:
@mlschicker I updated the run times and also added the same runs but using MPS simulation in a separate submission folder. It's interesting to see the comparisons.

mlschicker:
Hi @stephendiadamo , 

I tried to verify the your updated results. 
Unfortunately, the matrices you provide are not doubly stochastic as they are not convex combinations of permutation matrices - this is a hard requirement for this problem as we define it. 
Addressing numerical issues in other submissions, we accept a relative error of `1e-6` here. 

I see two options to resolve this issue:
- You re-add the classical post-processing to obtain a doubly stochastic matrix.
- You add a post-processing step that normalizes the weights. 

The first option would revert your results to have a lot of permutation matrices, whereas the second option will potentially worsen your error. 

Please let me know, which option you decide to use or if you see another option. 

Best, 
Max

stephendiadamo:
@mlschicker 

Thanks, I went with option 2 for this. The main thing is... I only ran 15 iterations for the QAOA routine, mainly to save on QPU resources. If we had a `1e-5` threshold, all submissions would pass, even up to the simulated B100 case and all the QPU runs as well. With `1e-6` it currently will not pass for all of the submissions. I wonder if we could loosen the tolerance... `1e-6` was chosen a bit arbitrarily, so perhaps we could raise it to `1e-5`? :)

mlschicker:
Hi @stephendiadamo , 

for the approximation, you can choose the tolerance as you wish. 
For being convex, i.e., the sum of the weights are 1, the tolerance should stay low. 
What do you think?

Best, 
Max

mlschicker:
Hi @stephendiadamo ,

I have updated your branch to have the new solution checker. 
One thing I notices is that the reported Frobenius norms do not coincide with the computed ones from the checker. 

I assume that the normalization of the solution happens after the run and hence the norm is different. 
Is this correct or is there another explanation?

Note that the submission checker by itself does not allow for approximate solutions. 
To allow this, change line `588` in `misc/ci/check_submission.py` to

```
return [str(binary), "--tolerance", "1e-7", str(instance_file), str(solution)]
```
Here, you can also adjust the tolerance. 
Please do not commit these changes.

Everything else looks good now. 

Best, 
Max

omearac:
@stephendiadamo anything else left to do here before the results are merged?

mlschicker:
Yes, can you explain the discrepancy in the norm?

Is this due to the normalization in the end?

stephendiadamo:
Hi, yes, what you assume is true. During execution the reported Frobenius norm error is calculated using unnormalized weights. In the post processing, the weights get rescaled.

mlschicker:
I checked your submission. 
The runs seem to have used different target approximations. 

You do not indicate in the success threshold, i.e., up to which approximation you run. 
Can you please add this information and adjust the number of successful runs accordingly? 

Best, 
Max

stephendiadamo:
@mlschicker Ok everything is updated.

mlschicker:
Hi @stephendiadamo , 

thank you for the update. 

I ran the submission checker again with tolerance `1e-6`. 

The following run are still misreported:

`20260803_PartitionedQAOA-EFCFW_DiAdamo/B64_64_1` 
- The solution has a norm of `1.039706e-6` as reported by the solution checker. This does not fulfill the requirement of a succesful run as you define it. 
- Moreover, it does not coincide with the norm you report in the objective time series (`1.0375738764648438e-06`). 


The simulator submission's solutions are OK with an approximation of `1e-5`.

For `B16_16_1`, you only report a single run even though we require at least 5 runs for any stochastic algorithm. 

If the data is available, please add all run's objective time series to the JSON files. 

And one more question: What is reported by `"Matrix_Residual": XXX` in the time series?
I would be nice, if you could add a small description in the README explaining all the fields you provide. 

Thank you!

Best, 
Max

stephendiadamo:
Hi Max, I did the following:

- I used the same tolerance of 1e-5 for everything
- I accidentally reran the B16_16_1 case in simulation after I pushed and overwrote the data only running 1 instance. I'll just remove the entry for now.
- I removed the Matrix_Residual data, it was mainly there to show the norm decreasing with scale of the problem, but it doesn't bring significantly unique information (it was just scale * norm). 
- The README norm and the time series norm differ due to a renormalization at the end of the process to meet QOBLIB's weight sum constraint

mlschicker:
Hi @stephendiadamo , 

thank you!

Best, 
Max

mlschicker:
The submission was verified by hand using `--tolerance 1e-5`. 
The objective time series is non-decreasing due to the method adding permutations. 

Everything looks good.

PR #32, opened 2026-08-05
E fcfw august 2026 pennington mohseni

victor-ibm:
E-FCFW run on quantum hardware for instances of size n=3 to n=10 (sparse) and n=3 to n=8 (dense). 

-  The instance file is valid and parses correctly with the problem's checker (check/ directory).
-  I believe the naming convention matches existing instances in that problem class.
-  The README is updated.
-  Best solution is provided. 
-  There is link to the paper. The code is open-source (link in the paper).


mlschicker:
Hi @victor-ibm , 

I checked the solutions and found a few inconsistencies:

- For `136 / 140` instances, the weights do not add up to 1. (scaled) Thus, the decomposition is not convex. If this check would be ignored, `68` instances pass. For the remaining `72` see the other comments. 
- I recomputed the error as indicated in the paper (`( 1 / n^2 ) * || D - X ||^2_F`) and found that these computations do not coincide with the reported errors. Indeed, `63` of the submitted solutions have an approximation error of `> 1e-6` even though they claim to be lower. 
- `9` instances report infeasible permutations: `Instance B9_9_5:     Permutation 2 is not a valid permutation: [0, 7, 0, 1, 0, 0, 9, 4, 0]`

Please check these points. 

Thank you!

Best, 
Max

omearac:
@victor-ibm this may be failing because of the exact verification. Max wrote a new checker here: #34

mlschicker:
Submission checked by hand:

All instances that report `successful` are withing `2e-6` in the specified norm:

$f(X) = (1/n^2) \\| X - D^* \\|^2_F$

Non-successful runs are marked correctly.

PR #34, opened 2026-08-07
Update Birkhoff checker to accept tolerance

mlschicker:
Update the birkhoff checker to accept approximate decomposition. 

By default, the decomposition must be exact and is checked entry-wise.
Passing `--tolerance {TOL}` to the checker enables a tolerance computed by:

$$\varepsilon = \frac{1}{n^2 \cdot \text{scale}^2} \sum_{i,j} \bigl(D_{ij} - \hat{D}_{ij}\bigr)^2  =  \frac{1}{n^2 \cdot \text{scale}^2}\bigl\Vert D_{ij} - \hat{D}_{ij}\bigr\Vert_F^2 $$

Note that the scale is there because solutions and instances are given as scaled integer matrices. 
Thus, the accumulation of the error happens in integer arithmetic (except the division in the end). 
The checker does not accept non-convex combinations of matrices, i.e., the weights must sum up to `scale`.


stefan-woerner:
👍

omearac:
Thanks Max, looks good to me too

PR #38, opened 2026-08-13
Add Quicopt v0.1 submission for 07-independentset (50 instances)

timbode:
Covers the full 50-instance independent-set suite, 20 independent runs per instance on a single CPU core.

Each instance directory holds the 30-column summary CSV and a solution file giving the 0/1 membership vector over the graph's vertices. The reported objective is the best set found across the 20 runs; the runtimes are means over those runs, as CONTRIBUTING requests. Every solution was verified against the instance graph before submission, and the whole tree passes qoblib-check-submission (50/50).

Optimality is not claimed anywhere: Optimality Bound is N/A on every row, and Algorithm Type is Stochastic.

This supersedes the 6-instance 20250627_Quicopt_Bode entry from the same submitter, which is left in place.


mlschicker:
Hi @timbode , 

I added the optional READMEs as they make the submission easier to view on GitHub. 

Would it be possible for you to provide objective time series and the time to solution for your runs?
It is okay, if that data is not available. 

Quick question on the choice of penalty: 
You use the canonical QUBO formulation with a penalty of 1. 
This means that the objective value will be the optimal one but the solution vector might possibly not infeasible. 

> As an example, consider the fully connected graph with two nodes; `O --- O`. 
The solutions `X --- O`, `O --- X`, and `X --- X` all give value 1, where the empty solution gives 0. 
Obviously, solution 3 is not feasible. 

Do you do some post-processing or do you always get feasible solution vectors from the algorithm?

Moreover, you mention you used a single core. 
Were the instances run concurrently?

Best, 
Max

timbode:
> Would it be possible for you to provide objective time series and the time to solution for your runs? It is okay, if that data is not available.

No, unfortunately I don't have these for this submission, sorry about that. But it should be close to the recorded budget anyway.

> Do you do some post-processing or do you always get feasible solution vectors from the algorithm?

Yes, this is guaranteed in the current set-up.

> Moreover, you mention you used a single core. Were the instances run concurrently?

Regarding concurrency: I run the 20 seeds for the 43 "small" instances on 20/128 cores. The 7 "large" ones are run sequentially on 1/128.

mlschicker:
> No, unfortunately I don't have these for this submission, sorry about that. But it should be close to the recorded budget anyway.

No problem.
Do you set the optimal value (if it is known) as a cutoff? 
Otherwise, what is the stopping criterion?

> Yes, this is guaranteed in the current set-up.

How is that guaranteed? 
Via post-processing?

> Regarding concurrency: I run the 20 seeds for the 43 "small" instances on 20/128 cores. The 7 "large" ones are run sequentially on 1/128.

OK.

**General Remark.** 
Could you describe your method in more detail please?
This should then also resolve my questions from above. 

Thank you!

Best, 
Max

timbode:
> Do you set the optimal value (if it is known) as a cutoff? Otherwise, what is the stopping criterion?

The shots are each deterministic, but their runtime is not known in advance. Repair then has a fixed time budget.

> How is that guaranteed? Via post-processing?

Exactly, I get proposals from the shots I run, and then do a pass of greedy repair, essentially.

> **General Remark.** Could you describe your method in more detail please? This should then also resolve my questions from above.

Not in more detail than above, sorry, since it's considered "commercial" now. Hope that's okay.

PR #39, opened 2026-08-13
[02-labs] Update best-known and optimal values from literature

mlschicker:
Updated best-known and optimal values for 02-labs. 

Information taken from https://github.com/borkob/git_labs/blob/master/results-2016/2004-Knauer-original.html. 

The solutions were independently verified by our solution checker. 

Added optional expected energy argument to the solution checker.


mlschicker:
@timbode

PR #42, opened 2026-08-16
Add Quicopt v0.2 submission for 02-labs (99 instances)

timbode:
One summary row and one solution file for every instance from labs002 to labs100, plus an objective time series at the 51 instances where every run's progress was recorded. Against the current contents of 02-labs this improves the best submitted objective at 54 instances and matches it at the other 45; every instance with a proven optimum is at the optimum.

Optimality Bound is N/A throughout. The solver is stochastic and proves no bound, so the entry makes no optimality claim even where the value coincides with a known optimum.

Runtimes are averages over the runs, as CONTRIBUTING asks. A run is one independent execution producing one candidate sequence; runs share a common pre-processing stage whose cost is divided among the runs that consumed it, so the per-run figures sum to what the complete workflow cost. Total Runtime is wall-clock and CPU Runtime CPU-seconds, and since the runs were executed concurrently the wall-clock includes contention on a memory-bandwidth-bound node -- it over-reports rather than under-reports. Every row's Remarks says so.

Time to Solution is read off the objective time series by the documented rule, the first entry reaching a run's final incumbent, averaged over the same runs as the other runtimes. Instances where a measured curve is not available for every run ship no series rather than a partial one.

Successful Runs counts the runs that reached the reported objective at epsilon = 0, so the rarer results declare themselves: labs100 reports 1 of 50.

Non-Zero Coefficients is the multilinear spin-domain expansion of sum_k C_k^2 after cancelling repeated indices with s^2 = 1, which gives 82075 terms of degree 2 and 4 at labs100.


mlschicker:
Hi @timbode , 

The submission format looks well.

I have a few questions about the results:

In the bigger runs, there are runtime discrepancies: 
```
| Total Runtime    |  426.733333 |
| Time to Solution |  317.291111 |
| CPU Runtime      | 3811.160000 |
```
How is the CPU runtime larger than the overall runtime?
Did you measure CPU time on each core individually? 
The time should be wall-clock time. 
If you want, feel free to add the actual CPU-time in the remarks as this is also an interesting measure.

Why are the objective time series only provided for some of the runs? 

What is the pre-processing step you mention and how much of the runtime does it actually consume? 

Thank you!

Best, 
Max

timbode:
> How is the CPU runtime larger than the overall runtime? Did you measure CPU time on each core individually? 

"Total Runtime" is wall, whereas "CPU Runtime" is summed over all cores and averaged over seeds that took part. The fractional number core number comes from global overhead that's accounted for as an average over seeds. Does that make sense?

> Why are the objective time series only provided for some of the runs?

For the easier case, the later stages of my algorithm aren't triggered because it already attains the optimum early.

So yes to this: 
> What it the pre-processing step you mention and how much of the runtime does it actually consume?

timbode:
ugh, the formatting is broken on my above comment - but I can't edit... GitHub says "Sorry, something went wrong."

mlschicker:
Hi @timbode ,

thank you for the clarifications. 

The time taken for the pre-processing step is only distributed across runs of the same instance, correct?
(Let $t_{pre}$ be the pre-processing time for an instance and $t_{run, i}$ be the runtimes for this instance, where $i \in [n]$. Then you average runtime for this instance is $(t_{pre} + \sum_i t_{run, i}) / n$, right?)

And do you know how long the preprocessing actually takes per instance?

Thanks!

Best, 
Max

timbode:
> The time taken for the pre-processing step is only distributed across runs of the same instance, correct? (Let t p r e be the pre-processing time for an instance and t r u n , i be the runtimes for this instance, where i ∈ [ n ] . Then you average runtime for this instance is ( t p r e + ∑ i t r u n , i ) / n , right?)

Yes, this is correct. 

> And do you know how long the preprocessing actually takes per instance?
 
 Here's an excerpt:
 
  | instance | pre-processing wall | pre-proc. CPU | runs sharing it | share of `Total Runtime` | share of `CPU Runtime` |
  | labs041  | 334 s | 20 040 s | 5 | 57.3 % | 87.0 % |
  | labs053 | 432 s | 21 600 s | 5 | 46.2 % | 78.2 % |
  | labs066 | 925 s (2 batches) | 50 570 s | 25 | 10.6 % | 49.3 % |
  | labs074 | 1023 s (2 batches) | 55 920 s | 45 | 5.3 % | 32.6 % |
  | labs090 | 645 s | 32 250 s | 5 | 24.5 % | 57.5 % |
  | labs099 | 682 s | 34 100 s | 5 | 21.0 % | 52.5 % |
  | labs100 | 1220 s (2 batches) | 73 200 s | 50 | 4.2 % | 18.1 % |
  
  It's a proper part of the computation, don't think it makes sense to break it out unless going into all of the details, which I can't anyway.

mlschicker:
Hi @timbode , 

We decided that in this case you should report the minimum runtime, i.e., `t_pre + t_alg` as otherwise the pre-processing runtime will have decreasing effect on the total runtime which it should not. 

Moreover, in the remarks, please report the average value of `t_alg` without `t_pre` and report `t_pre` additionally. 

Thank you!

Best, 
Max

timbode:
Hi @mlschicker,

yeah, makes sense! I'll change it. I will report wall time as the actual "waiting time" on the clock given the cores, which should indeed count the "pre-processing" (a misnomer since it's a key part of the overall algorithm) separately for every seed. I'm updating the PR.

timbode:
I actually have a small update that would _improve_ the runtimes - should I just push fresh results to the forked branch?

mlschicker:
Hi, yes. Feel free to update the results.

mlschicker:
I converted the PR to a draft. 
When you're ready, hit the `Ready for review` button and I will take another look. 

Thanks!

mlschicker:
Hi @timbode , 

thanks for the update!

I am a bit surprised by the following numbers: 

| Field | Value|
|---|---|
| Total Runtime | 1343.760000 |
| Time to Solution | 1073.687500 |
| CPU Runtime | 29932.560000 |
| GPU Runtime | 0 |
| QPU Runtime | 0 |
| Other HW Runtime | 0 |
| ====== |  |
| Remarks | A run has two parts. Work done for that run alone takes 939 s wall and 5633 s CPU on average. A preparation step, shared by all 40 runs of this instance, takes 405 s wall and 24300 s CPU.|

The total runtime is now the best runtime including pre-processing, correct?
The average is quite close. 

Best, 
Max

timbode:
@mlschicker Wait, I'm not following. Which average is close to what? Compared to the previous numbers, there has also been an algo update.

mlschicker:
The best runtime numbers are very close to the average. 
So all of the jobs took about the same time?

timbode:
Not that close, are they? Anyhow, that's to be expected? They're just different seeds doing the same stuff.

PR #43, opened 2026-08-17
Add exact Birkhoff benchmark runs

mnn31:
Adds deterministic exact Birkhoff decompositions for all 375 benchmark
instances. Fifteen of them, the `B64_4096_*` and `B100_10000_*` families, had no
feasible solution on record at all.

The solver picks a positive perfect matching with the largest possible minimum
residual, breaking ties by first maximising the number of residual entries it
zeroes out and then minimising the matching residual sum. Where that still
leaves several equally optimal matchings, the choice is made lexicographically:
rows are fixed in order, each taking the smallest column that still admits an
optimal completion. That pins the result independently of which optimal matching
the LAP implementation happens to return. All arithmetic is on the scaled
integers, so every decomposition reconstructs its matrix exactly and passes
`03-birkhoff/check` at the default tolerance of 0.

Each instance directory carries the decomposition, an objective time series, the
measured solve time and hardware, and a link to the solver.

Running `misc/ci/update_bkv.py --check` credits 276 instances to this
submission. An earlier version of this description said 296, which was wrong:
that was the total number of changed rows in the table, and 21 of them belong to
other people's submissions and would change regardless of this PR. The count
moved from 275 to 276 when the lexicographic tie-breaking was added.

Solver and reproduction notes: https://github.com/mnn31/qoblib-birkhoff

Known issue: `check_submission.py` fails on the objective time series. The number
of matrices grows as the decomposition is built, so the series is non-decreasing
rather than non-increasing. That is the point raised in the thread. Everything
else passes.


mlschicker:
Hi @mnn31 , 

I have added the optional READMEs to increase legibility in the repository. 

It seems that you have pushed your experiments folder by accident. 
Please remove it from the PR. 

Would it be possible to provide the objective trajectory here?
This would include: Time, Number of Matrices and Approximation (as the normalized squared Frobenius norm).
$f(X) = \frac{1}{n^2} \Vert D - X \Vert^2_F$
This way, we could compare it to the other submissions. 

Moreover, would it be possible to run the other instances too?
I think this would be valuable. 

Thank you!

Best, 
Max

mnn31:
Thanks for the helpful review. I have updated the PR:

- Removed the experiments directory from this repository.
- Moved the solver and reproduction notes to https://github.com/mnn31/qoblib-birkhoff.
- Added a trajectory JSON file for each submitted instance. Each point records Time, Number of Matrices, and the normalized squared Frobenius residual.

I am also running the remaining Birkhoff instances now and will add any checked results in a follow-up update.

mlschicker:
Hi @mnn31 , 

thank you for the quick update!

Please rename the objective time series according to the conventions, i.e., `<INSTANCE_NAME>_objective_time_series.json`.

Moreover, in the objective time series, rename `Number of Matrices` to `Incumbent`. 
This way, the automatic parsers will accept it. 

Looking forward to receiving the remaining runs!

Best, 
Max

mnn31:
Updated in ec22240f. The files now use `<INSTANCE_NAME>_objective_time_series.json`, with a nested run list and the Birkhoff convention `Time`, `Incumbent`, and `Error` (the normalized squared Frobenius residual). The remaining-instance batch has completed; I am validating the results now and will follow up with checked additions.

mnn31:
The remaining runs are now included in fcdd374c: 360 additional exact decompositions with solution files, summaries, and objective time series. The best-known-value update records 296 newly established or improved exact values. Validation completed with strict submission checks (15/15 and 360/360), an independent exact reconstruction pass over all 375 runs, and the BKV consistency check.

mlschicker:
Hi @mnn31 , 

thank you for the update. 

Please do not change the CI or the solutions directory. 
The contents of the solutions directory is automatically updated upon merge. 

The Validation fails at the moment as the objective time series should technically be non-increasing but is non-decreasing for your method. 
This is OK and will be handled differently in the future. 

Thanks!

Best, 
Max

mnn31:
Thanks Max, both reverted.

`03-birkhoff/solutions` is back to main, the 276 modified files and the 15 new
ones. `misc/ci/check_submission.py` is back to main too, I should not have
touched the validator to make my own submission pass. The diff is now 1141
files, all under `03-birkhoff/submissions`.

On the validation failure, that is the monotonicity point you already spotted.
The only remaining errors are the objective time series being non-decreasing,
because the number of matrices grows as the decomposition is built. Two options
if you want it green before the handling changes:

- report the normalised squared Frobenius residual as `Incumbent` instead, which
  does decrease monotonically and reaches 0 at the end, and keep the matrix count
  under a separate key, or
- drop the series entirely.

I have left it as `Incumbent` = number of matrices with `Error` = residual for
now, since that is what you asked for earlier. Say which you prefer and I will
switch it.

One correction to the description. `update_bkv.py --check` credits 275 instances
to this submission, not the 296 I claimed. 296 was the total number of changed
rows in the table, and 21 of those belong to other submissions and would move
regardless of this PR. I have fixed the description.

mlschicker:
Hi @mnn31 , 

thank you for the update. 

Is there a reason, the submission is split in two?
From the description, the method is the same. 
So I think one submission would make more sense. 

Moreover, it is fine that the check fails on this. 
This is a known issue which will be updated in the future.

Best, 
Max

mnn31:
Merged, thanks. No good reason for the split, it was just the order I ran them
in. Everything is now under `20260817_BirkhoffMaxMin_Gupta`, 375 instances, and
the root README covers the whole set with a per family breakdown. Also generated
the per instance READMEs for the 360 that were missing them.

Good to hear the monotonicity failure is expected. I left the series as
`Incumbent` = number of matrices with `Error` = normalised squared Frobenius
residual, as you asked. The offer stands if you would rather have the residual
as the incumbent later, it decreases monotonically to 0 so it would pass the
current validator unchanged.

mlschicker:
Hi @mnn31 , 

thank you for the update and reverting the CI to its original form. 

You claim that the method is deterministic and you use a library to compute matchings which are central. 
Are you sure the tie-breaking is handled deterministically? 

Best, 
Max

mnn31:
Good question, and the honest answer is more nuanced than the CSV field suggests.

What I can show:

- There is no randomness in the solver. No RNG, no seeding, no wall clock in any
  decision, and the only iteration over a mapping is over the instance payload,
  which is insertion ordered. The single algorithmic choice is
  `scipy.optimize.linear_sum_assignment` on an integer cost matrix.
- Reruns are bit identical. Decomposing the same instance collection twice in
  the same process and hashing the full weight and permutation lists gives the
  same digest, on 20 instances of size 16 and 32.

What I cannot claim, and where you are right to push:

- The tie breaking is not pinned by the method. The stated rules, maximise the
  smallest residual, then maximise eliminated entries, then minimise the
  matching residual sum, are all folded into the cost matrix, but they do not
  always single out one matching. When several assignments achieve the same
  optimal cost, which one comes back is decided by the traversal order inside
  scipy's LAP implementation, not by anything the algorithm specifies.
- I measured how often that actually happens. Taking the cost matrix at each
  step, solving it, then forbidding each assigned cell in turn and re-solving,
  an alternative equally optimal matching exists at 2 of 203 steps probed, about
  1 percent, across 25 instances.

So the results are reproducible on a fixed scipy build, which is what the
Deterministic field was meant to convey, but they are not canonical. A different
scipy version or a different LAP implementation could return a different matching
at one of those 1 percent of steps and end up with a different number of terms.

I would rather not leave that implicit. Two options, your call:

1. I reword the Remarks to state exactly the above, that the method is
   deterministic and reproducible on a fixed scipy build but delegates tie
   breaking among equally optimal matchings to the LAP implementation, and
   record the scipy version in the hardware or remarks field.
2. I pin it properly. Detect the tied steps, which are rare, and resolve them
   with an explicit lexicographic rule so the output is implementation
   independent, then rerun all 375. This would change a small number of the
   reported term counts, in either direction.

I lean towards 2 since it makes the claim true without qualification, but it
does mean new numbers, so tell me which you prefer.

scipy version used for the submitted runs: 1.17.1.

mlschicker:
Hi @mnn31 , 

I prefer the lexicographic tie-breaking. 

Best, 
Max

mnn31:
Done, lexicographic it is.

The rule: among equally optimal assignments, fix rows in order and give each row
the smallest column that still admits an optimal completion. Costs are integers
so the optimality test is exact. That singles out one assignment regardless of
which optimal matching the LAP returns, so the output no longer depends on the
scipy version.

It is applied to the winning threshold's cost matrix at every step, and to the
largest_weight policy as well. The binary search itself only needs feasibility,
so it is left alone.

Reran all 375. Compared to the previous submission:

- 33 instances improved
- 38 got worse
- 304 unchanged
- net +6 matrices across the whole set

Which is what you would expect given how rare the ties are. When I measured it
earlier, only about 1 percent of assignment steps had an alternative equally
optimal matching, so most instances could not move.

Worth being explicit that this is a real trade. The submitted values are now
slightly worse in aggregate than the ones I sent before. I think that is the
right way round, since the previous numbers were partly an artifact of scipy's
traversal order and were not reproducible against a different LAP
implementation, whereas these are.

Cost is small, the whole set reruns in a few minutes.

All 375 solutions pass 03-birkhoff/check. The submission check still fails only
on the time series monotonicity, which you have said is expected for a
constructive method. Per instance READMEs regenerated.

mnn31:
Unrelated to this PR, but I ran into it while checking why update_bkv credits me
for B100_100_1 rather than someone with a better value, so passing it on.

`20260805_PartitionedQAOA-EFCFW-Simulator_DiAdamo` reports 60 for B100_100_1,
which would be far better than anything else on that instance. It does not
verify:

```
$ check_birkhoff instances/qbench_100_sparse.json \
    submissions/20260805_PartitionedQAOA-EFCFW-Simulator_DiAdamo/B100_100_1/B100_100_1_solution.json
  Instance B100_100_1:     Reconstruction mismatch at index 0: expected 61629, got 0
INFEASIBLE: 1 of 1 instances failed
```

The same file checks out fine for my own entry on that instance:

```
  Instance B100_100_1:     Valid decomposition with 283 permutation matrices
VALID: All 1 instances verified successfully
```

So update_bkv is behaving correctly, it is not counting a value whose solution
does not reconstruct, and my attribution is not resting on that submission being
overlooked. But the submissions listing still shows 60 for that instance, which
is worth knowing about since it is merged and reads as a record on the site. I
have not checked their other instances.

Happy to open a separate issue if you would rather track it there.

mlschicker:
Hi @mnn31 , 

thank you for the update!

As for the the other submission:
This is a submission for the approximate problem. 
There is a threshold on the normalized squared Frobenius norm of the difference. 
At the moment, only exact decompositions are credited. 
Thus, it does not show up. 

Best, 
Max

mlschicker:
Quick follow up question:
How do you count the number of variables, non-zero coefficients, ...?

mlschicker:
Hi @mnn31 , 

please answer the last open question. 
Then, we can continue with the review. 

Thank you. 

Best, 
Max

mnn31:
Sorry for the slow reply on this one.

First, thanks for the explanation on the other submission. That resolves it
completely: it targets the approximate problem with a threshold on the
normalised squared Frobenius norm, and only exact decompositions are credited,
so the checker rejecting it under the exact criterion is the system working as
intended rather than a stale record. My report was based on the wrong
assumption, apologies for the noise.

On the counting, the honest answer is that those fields are weaker than they
look, and one of them is simply wrong.

What the script currently emits, for a decomposition with K terms on an n by n
matrix:

| field | formula | B10_100_1, n=10, K=52 |
| :--- | :--- | ---: |
| # Decision Variables | K(n+1) | 572 |
| # Binary Variables | K·n | 520 |
| # Integer Variables | K | 52 |
| # Continuous Variables | 0 | 0 |
| # Non-Zero Coefficients | 2n | 20 |

Two problems with that.

The non-zero count is wrong. 2n does not depend on K at all, so it reports the
same 20 whether the decomposition has 5 terms or 500. It looks like it was meant
to be the number of row and column sum constraints on the doubly stochastic
matrix, which is 2n, and got put in the wrong column.

The binary count is a permutation encoding, not binaries. K·n counts one entry
per row of each permutation, holding the chosen column. Those are not binary
variables. A permutation matrix written with binary variables in the usual
assignment form needs n² per term, so K·n² = 5200 here, not 520.

Underneath both is that this method never builds an optimisation model. It is
constructive: repeatedly pick a positive perfect matching, subtract, repeat. So
there is no model whose variables and coefficients these fields could describe.
What they are actually describing is the size of the emitted decomposition.

Two ways to fix it, and I would rather you pick than guess:

1. Report the size of the emitted solution and say so, since that is what the
   numbers really are. K permutations of length n plus K weights, and leave the
   coefficient fields as N/A because no model is built.
2. Report the assignment formulation the problem would have if it were modelled
   as a MIP: K·n² binaries for the permutation entries, K integer weights, row
   and column constraints per term, plus the reconstruction constraints. Worth
   noting the reconstruction constraint is bilinear in w and x, so it is not a
   linear model without linearisation, and any coefficient count depends on how
   that is done.

I lean towards 1, since it does not claim a model that was never solved, but 2
is more comparable with the MIP based submissions. Tell me which you prefer and
I will regenerate all 375 summaries accordingly. Either way the non-zero field
gets corrected.

mlschicker:
Option 1 sounds better to me.

mnn31:
Done, option 1, and sorry this took a week.

The summaries now describe the size of the emitted decomposition:

| field | value |
| :--- | :--- |
| # Decision Variables | K(n+1), the K weights plus K permutations of length n |
| # Binary Variables | 0 |
| # Integer Variables | K(n+1) |
| # Continuous Variables | 0 |
| # Non-Zero Coefficients | N/A |
| Coefficients Type | N/A |
| Coefficients Range | N/A |

Modeling Approach and Remarks both say plainly that no optimisation model is
built, so nobody reading one directory has to infer it. The old non-zero count
is gone.

Your point on the network PR about not knowing which runs shared the CPU applies
here too, since these were produced with six parallel workers. So I reran all
375 strictly one at a time on an otherwise idle machine before regenerating.
Every objective is identical to the previous version, which is what a
deterministic method should give. Total runtime went from 19.9 minutes to 13.3
minutes, a factor of 1.50, which lines up with the 1.52 I measured for the
network search, so the earlier runtimes were inflated by about that much. The
hardware field now states the one-at-a-time condition.

All 375 solutions pass 03-birkhoff/check, update_bkv still credits 276, and the
only remaining validation failure is the time series monotonicity you mentioned.
The solver repository has the same change.

PR #44, opened 2026-08-17
Add ChainDP submission for 06-portfolio (160 instances)

mnn31:
exact dynamic program over per-period portfolios. the objective only couples
periods through the rebalancing term, and the model charges no rebalance into
the last period, so it is a chain and a forward dp solves it exactly.

what is here, 160 instances:

- a003, a004 and a005, 96 instances, all listed open with no feasible solution
  on record. now solved to proven optimality.
- a010_t10 and a010_t15, 64 instances. these all match the published values so
  nothing improves, but they are now proven optimal rather than best known. 50
  of the 64 were listed best known.

no lp or qubo model files are shipped for the a003 to a005 families, which is
probably why they were never solved. the instance data was always there.

the objective is reimplemented in exact rational arithmetic with zimpl rounding
and was validated against the shipped a010 reference solutions before anything
here was produced. it reproduces their published values exactly, including the
ones marked optimal. all 160 solutions pass 06-portfolio/check, and
check_submission.py reports 160/160.

one thing you may want to look at separately. solutions/po_a003_t02_orig.bst.sol
has a comment line reading "Original (raw QUBO, incl. penalty) objective value =
-204.0878963861868" and a canonical header of "objective -1595". the reference
loader picks up the comment value, so the best known table lists a003_t02_orig
at -204.09, which is a raw qubo cost including a penalty term rather than a
model objective. that key also has no _bXXX_lY suffix, so it swallows all eight
lambda variants of a003_t02_orig in update_bkv and they do not get separate
rows. the other 88 tiny instances are unaffected. happy to split that out into
an issue if you prefer.

code and full method write up: https://github.com/mnn31/qoblib-net


mlschicker:
Hi @mnn31 , 

Could you please add the READMEs to the submission?
This way, it is easier to browse on GitHub. 
Just run the submission checker with `--generate-readme`. 

Moreover, is it possible to provide the objective time series? 

As for the references repository:
It also includes results for the network problem and the main README mainly describes this. 
Could you split the two submissions cleanly so that the reference is clearer?

Thank you!

Best, 
Max

mnn31:
Hi Max, thanks for looking at this.

All three done.

READMEs: regenerated the submission with `--generate-readme`, so every instance
directory now has one, plus the root README.

Objective time series: added one per instance. The method is an exact dynamic
program, not anytime, so it produces no incumbent until it finishes and each
series is the single point it returns. Happy to drop them if a one point series
is more noise than signal for your TTS plots, but I would rather report the
shape of the algorithm honestly than fabricate a curve.

Reference repository: split. It is now
https://github.com/mnn31/qoblib-solvers with a `portfolio/` and a `network/`
directory, each with its own code, results and README, and the top level README
only points at the two. The Reference field of this submission now points at
`tree/main/portfolio` rather than the repo root. The network work is a separate
submission and is not mixed in here.

check_submission.py still reports 160/160 after the changes.

mlschicker:
Hi @mnn31 , 

thank you for the update. 

In the remarks of the runs you note that this is not an anytime algorithm and mention that the objective time series thus only consists of a single value. 
The remark is enough. 
You can remove the objective time series in this case as it is unnecessary. 

Thank you!

Best, 
Max

mnn31:
Removed, thanks. The remark in each summary CSV still explains why there is no trajectory for this method.

mlschicker:
Hi @mnn31 , 

thank you for the update!

Two more things that need clarification:
- What CPU was this run on? Please provide the exact model. 
- Please explain the rounding you used independently in the main readme. 

Thank you!

Best, 
Max

mnn31:
Both done.

CPU: Apple M3 Pro, Mac15,6, 11 cores split 5 performance and 6 efficiency, 18 GB
unified memory, macOS 26.3, arm64. Updated in all 160 summaries and the
regenerated READMEs. Worth noting the core split since the larger a010 runs were
spread across workers, so some of them landed on efficiency cores and the
reported runtimes are not all from performance cores.

Rounding: added a Rounding section to the submission README. Short version, the
objective is never evaluated in floating point. Prices and covariances are
parsed into exact rationals and unit prices are rebased exactly as
p[i,t] = raw_p[i,t] * unit / raw_p[i,0] before anything is rounded. Zimpl's
round is half away from zero, so 0.5 goes to 1 and -0.5 goes to -1, which is not
what IEEE or Python's round does; using banker's rounding shifts individual
coefficients by a unit. Rounding is applied once per model coefficient exactly
where the reference model applies it and never to a running total, and the
README lists the expression rounded for each one. After that every coefficient
is an integer, so the dynamic program is integer arithmetic throughout with no
accumulation error.

That is also why the a010 families are in this submission at all. They do not
improve anything, but reproducing all 64 published values exactly is the check
that the rounding matches your model term by term.

Still 160/160.

mlschicker:
Hi @mnn31 , 

thank you for the update. 
I would argue that including the results for the `a010` family are worth including as they prove optimality on the previously known values - this is very valuable information. 

Why did you run some of the instances on performance and some on efficiency cores?
How big is the impact of this in your experience? 
It would be important to note if there are big discrepancies in the runtimes due to this. 

Best, 
Max

mnn31:
Agreed on a010, they stay.

On the cores, the honest answer is that I did not choose it. I ran eight
concurrent workers on a machine with 5 performance and 6 efficiency cores, so at
least three solves were always on efficiency cores. It was a side effect of the
parallelism, not a decision, and you are right that it makes the runtimes hard
to compare.

You asked me to note the discrepancy. It turned out big enough that noting it
was not really enough, so I reran the timings properly. Two effects, measured
separately on this machine:

Core type, same instance and lambda, idle machine, median of three with a warm
up discarded:

| condition | median solve |
| :--- | ---: |
| performance cores | 75.2 s |
| efficiency cores | 118.2 s |

so 1.57x from core type alone.

Contention, all 64 a010 instances:

| condition | total |
| :--- | ---: |
| eight concurrent workers | 238.5 min |
| single process, idle machine | 105.1 min |

so 2.27x end to end. The gap between 1.57x and 2.27x is memory and cache
pressure from eight solves running at once, not core placement.

The submission now carries the single process numbers, and the README explains
where they come from. All 160 objectives are bit identical between the two runs,
so only the timings moved, which is also a decent determinism check on the
method.

For the small families the same effect is 1.89x, 6.0 s against 3.2 s, though at
that scale it hardly matters.

Still 160/160.

mlschicker:
Hi @mnn31 , 

The runs should be run exclusively on your machine if possible. 

Did you do this?

Thank you. 

Best, 
Max

mnn31:
Yes. The numbers currently in the submission are from a rerun with a single
process and nothing else on the machine.

To be precise about the history, since the first version of this submission was
not measured that way. The original timings came from a run with eight
concurrent workers, which is what prompted your question about performance
versus efficiency cores. After that exchange I reran every instance
single-process on an otherwise idle machine and replaced the runtime column
entirely. The a010 families went from 238.5 minutes contended to 105.1 minutes
exclusive, a factor of 2.27, with all 160 objectives bit identical between the
two runs.

The submission README documents this, and the hardware field says the runtimes
come from a single-process rerun rather than from the parallel one.

So the values you are looking at now satisfy what you are asking for. The
contended numbers are gone.

PR #45, opened 2026-08-18
Add ParallelTempering submission for 08-network (20 instances, 8 improving)

mnn31:
Improves the best known value on network13 through network23, eleven of the
fourteen instances that were not already proven optimal. Biggest is network17 at
5.6 percent, 460182 to 434429.

| instance | published | this submission | improvement |
| :--- | ---: | ---: | ---: |
| network13 | 304116 | 299715 | 1.45% |
| network14 | 350173 | 342455 | 2.20% |
| network15 | 383000 | 370967 | 3.14% |
| network16 | 409067 | 401000 | 1.97% |
| network17 | 460182 | 434429 | 5.60% |
| network18 | 481950 | 474643 | 1.52% |
| network19 | 514625 | 497000 | 3.42% |
| network20 | 548536 | 526653 | 3.99% |
| network21 | 593000 | 570022 | 3.87% |
| network22 | 647594 | 621810 | 3.98% |
| network23 | 686453 | 660000 | 3.85% |

The problem separates. Picking the digraph is the hard part, but once it is
fixed the routing is a min congestion multicommodity flow, which is an LP. So
the search only moves over topologies and scores each one exactly.

Two things make that work. On all 20 published solutions the ceiling of the
routing LP equals the published objective exactly, so the LP value is a sound
search energy and integrality only has to be restored at the end. And a 2 in 2
out digraph is the union of two fixed point free permutations that disagree
everywhere, so the 2 exchange (a->b),(c->d) => (a->d),(c->b) preserves every
degree and every proposal is feasible by construction.

Parallel tempering, eight replicas, geometric temperature ladder, replica
exchange every 40 proposals, half the replicas seeded from the published
reference topology and half random. Five independent runs per instance, seeds 0
to 4, single core each, 20 minutes per run on a laptop. The integral routing is
recovered once at the end by re-solving the same model with integrality on the
flow variables.

Objective time series included, one entry per incumbent improvement per run.
Exactly one run per instance reaches the best value, so a single run understates
the method. Time to Solution in the CSVs is the average over runs of the moment
each run last improved.

network11 and network12 were run five times each and matched their published
values every time without beating them. I left network24 alone since its
continuous flow version is MIPLIB's dano3mip.

All eleven pass 08-network/check and check_submission.py reports 11/11. Per
instance READMEs generated with --generate-readme. Nothing outside
08-network/submissions is touched.

Code and write up: https://github.com/mnn31/qoblib-solvers/tree/main/network


mlschicker:
Hi @mnn31 , 

thank you for you submission. 
As it stands, the submission cannot be merged.

It is not allowed to use the published best-known (or optimal) solutions as a starting solution. 
Otherwise, every method could easily reach the best-known solution which would be misleading. 

Please fix your workflow to **not** use the published best-known values.

Of course, you can split the method into two phases by first computing a solution using any MIP solver using, e.g., the provided model, and then running the proposed PT on the given solution. 
In this case, the two phases must be explicitly documented. 

Best, 
Max

mnn31:
You are right, and thanks for catching it. Seeding half the replicas from the
published topology means the run starts at the incumbent and can only report
something at least as good, so the numbers do not measure the search. That
should not have gone out.

I have removed the reference seeding from the code itself rather than just from
the run, so it cannot come back by accident. The sweep no longer reads the
solutions directory at all, every replica now starts from an independently
sampled random 2 in 2 out topology, and the same warm start options are gone
from the single instance driver.

Re-running now from random starts only, five seeds per instance, with a longer
budget since the search has further to travel. I will post the results either
way. If the improvements do not survive without the warm start I will say so and
withdraw the affected instances rather than quietly reporting a smaller set.

Marking this as a draft until then so it is not sitting in your queue as
mergeable.

On the two phase option, I would rather first see what the method does on its
own from random starts. If a construction phase turns out to be needed I will
build one that only uses the instance data, document both phases separately, and
report the phase one value as well so the contribution of each is visible.

mnn31:
Re-ran everything from random starts. The search no longer reads the solutions
directory at all, every replica begins at an independently sampled random 2 in 2
out topology, and the warm start options are gone from the code rather than just
unused. Five runs per instance, seeds 0 to 4, 40 minutes each on one core.

You were right to push on this. Three of the eleven instances I claimed do not
survive, and one more came out worse than the published value. I have withdrawn
all four. The submission is now eight instances:

| instance | published | this submission | improvement |
| :--- | ---: | ---: | ---: |
| network13 | 304116 | 300620 | 1.15% |
| network14 | 350173 | 343091 | 2.02% |
| network15 | 383000 | 370870 | 3.17% |
| network16 | 409067 | 400072 | 2.20% |
| network17 | 460182 | 437000 | 5.04% |
| network18 | 481950 | 476758 | 1.08% |
| network19 | 514625 | 497473 | 3.33% |
| network20 | 548536 | 528528 | 3.65% |

Withdrawn, for the record rather than quietly dropped:

| instance | published | best from random starts | |
| :--- | ---: | ---: | :--- |
| network12 | 276474 | 277527 | worse |
| network21 | 593000 | 627462 | worse |
| network22 | 647594 | 686514 | worse |
| network23 | 686453 | 720834 | worse |

The pattern is clean and worth stating: the method holds up to about 20 nodes and
falls off a cliff after that. Warm started, 21 to 23 looked like 4 percent
improvements; cold, they are 5 to 6 percent worse than published. That gap was
the seed, not the search, which is exactly what you were warning about.

network11 matched its published value in all five runs without beating it.
network24 was left alone.

One fix along the way. The declared objective used to be the solver's z, which
the model only bounds from below, so an untight z could disagree with the
checker's recomputed maximum arc load. It is now recomputed from the flows.

All eight pass 08-network/check and check_submission.py reports 8/8. Marking
ready for review.

mlschicker:
Hi @mnn31 , 

thank you for the update. 

It is also interesting to have the runs that performed worse than the currently best-known solution in the submission. 
This way, it is easier to compare methods in the long run. 

Could you please add the runs that performed worse than the best-known? 

Best, 
Max

mnn31:
Added, thanks. The submission now carries all thirteen instances that were run,
not just the eight that win.

| instance | published | this method | |
| :--- | ---: | ---: | :--- |
| network13 | 304116 | 300620 | improves |
| network14 | 350173 | 343091 | improves |
| network15 | 383000 | 370870 | improves |
| network16 | 409067 | 400072 | improves |
| network17 | 460182 | 437000 | improves |
| network18 | 481950 | 476758 | improves |
| network19 | 514625 | 497473 | improves |
| network20 | 548536 | 528528 | improves |
| network11 | 238334 | 238334 | matches, does not beat |
| network12 | 276474 | 277527 | worse |
| network21 | 593000 | 627462 | worse |
| network22 | 647594 | 686514 | worse |
| network23 | 686453 | 720834 | worse |

Each non-improving run says so in its Remarks, so nobody reading a single
directory has to infer it. update_bkv still credits only the eight, as it
should.

The five that do not win are the more informative half, I think. The method
holds to about 20 nodes and then falls off: at 21 to 23 it lands 5 to 6 percent
above the published values rather than below. That is the same boundary where
the warm start you caught was doing all the work, so the two observations line
up.

All thirteen pass 08-network/check, check_submission.py reports 13/13. I also
retitled the PR since eight was no longer the right number.

mlschicker:
Hi @mnn31 , 

if possible, please run your code on **all** of the instances. 
It should be fairly quick for the small ones. 

This way, it is easier to see how well the method performs overall. 
It is not only about finding the best objective but also the runtime. 

Best, 
Max

mnn31:
Running now. I am extending the sweep to network05 through network24, using the
same protocol as the existing entries so the runtimes stay comparable: five
independent runs per instance, random starts only, eight replicas, 2400 seconds
per run, one core per run.

That adds the six instances that are already proven optimal, network05 to
network10, plus network24, to the thirteen already in the PR. Agreed that the
optimal ones are the useful ones for judging the method, since there the target
is known and the only question is whether it gets there and how fast. The time
series will show time to solution for each.

I will post the full table when it finishes, including network24, which I expect
to lose on given its continuous flow version is MIPLIB's dano3mip.

mnn31:
Done, all twenty instances, same protocol throughout: five independent runs,
random starts only, eight replicas, 2400 s per run, one core per run.

| instance | published | this method | | mean TTS | runs reaching it |
| :--- | ---: | ---: | :--- | ---: | :--- |
| network05 | 65500 | 65500 | matches proven optimum | 0.1 s | 5/5 |
| network06 | 101000 | 101000 | matches proven optimum | 0.6 s | 5/5 |
| network07 | 142400 | 142400 | matches proven optimum | 6.4 s | 5/5 |
| network08 | 170231 | 170231 | matches proven optimum | 8.0 s | 5/5 |
| network09 | 196750 | 196750 | matches proven optimum | 33.4 s | 5/5 |
| network10 | 210800 | 210800 | matches proven optimum | 386.8 s | 5/5 |
| network11 | 238334 | 238334 | matches | 890.8 s | 2/5 |
| network12 | 276474 | 277527 | +0.4% | | 1/5 |
| network13 | 304116 | 300620 | improves 1.15% | | 1/5 |
| network14 | 350173 | 343091 | improves 2.02% | | 1/5 |
| network15 | 383000 | 370870 | improves 3.17% | | 1/5 |
| network16 | 409067 | 400072 | improves 2.20% | | 1/5 |
| network17 | 460182 | 437000 | improves 5.04% | | 1/5 |
| network18 | 481950 | 476758 | improves 1.08% | | 1/5 |
| network19 | 514625 | 497473 | improves 3.33% | | 1/5 |
| network20 | 548536 | 528528 | improves 3.65% | | 1/5 |
| network21 | 593000 | 627462 | +5.8% | | 1/5 |
| network22 | 647594 | 686514 | +6.0% | | 1/5 |
| network23 | 686453 | 720834 | +5.0% | | 1/5 |
| network24 | 663688 | 702028 | +5.8% | | 1/5 |

You were right that the solved instances are the informative ones. The method
reaches the proven optimum on all six, in five runs out of five, and the time to
solution grows cleanly with size: 0.1, 0.6, 6.4, 8.0, 33.4, 386.8 seconds for
n = 5 to 10. That is roughly an order of magnitude every two nodes, and it is a
much better characterisation of the method than the improvement count is.

Extrapolating that curve is also what the rest of the table shows. At n = 11 it
still lands on the known value but only in 2 runs of 5 and takes 890 s. From
n = 21 up it is 5 to 6 percent above the published values. So the working range
of this method on this problem is roughly n <= 20 at this budget, and the eight
improvements sit inside it rather than being spread evenly.

network24 is included and loses by 5.8 percent, which is unsurprising given its
continuous flow version is dano3mip.

All 20 pass 08-network/check, check_submission.py reports 20/20, update_bkv
credits the 8 improvements only. Per instance READMEs regenerated.

03-birkhoff/instances/qbench_100_dense.json
mlschicker, line 1: Do **not** delete any instances in the repository!

08-network/submissions/20260818_ParallelTempering_Gupta/network05/README.md
```diff
@@ -0,0 +1,42 @@
...
+| Hardware Specifications | Apple M3 Pro (Mac15,6), 11 cores (5 performance + 6 efficiency), 18 GB unified memory, macOS 26.3, arm64; one core per run |
+| ====== |  |
+| Total Runtime | 2400.0 |
+| Time to Solution | 0.1 |
+| CPU Runtime | 2400.0 |
+| GPU Runtime | N/A |
+| QPU Runtime | N/A |
+| Other HW Runtime | N/A |
+| ====== |  |
+| Remarks | Heuristic, so the optimality bound is left as N/A. Runtimes are the average over the independent runs, single core each, queueing excluded. Time to solution is the average over runs of the moment each run last improved its incumbent. This run does not reach the published best-known value and is included so the method stays comparable over time rather than only appearing where it wins. Successful runs are those reaching this method's own best value. The declared objective is recomputed from the flows rather than taken from the solver's z variable, which the model only bounds from below. Verified with 08-network/check. |
```
mlschicker, line 42: Were the runs exclusive on the CPU or run simultaneously?

03-birkhoff/instances/qbench_100_dense.json
mnn31: Fixed, sorry about that. It was not intentional: the file went missing from my
working tree and a `git add -A` in the following commit recorded the deletion. I
have restored it from the base commit and verified `git diff` against upstream
for all of `03-birkhoff/` is now empty. The PR touches only
`08-network/submissions/` again.

08-network/submissions/20260818_ParallelTempering_Gupta/network05/README.md
mnn31: Simultaneously, and thank you for asking, because it means these numbers are not
comparable to the ones in my portfolio submission.

The sweep ran eight concurrent workers on an 11 core machine, so every run in
this PR shared the CPU. I measured what that costs on this workload:

| condition | LP evaluations per second, network17 |
| :--- | ---: |
| eight concurrent workers, as submitted | 104.0 |
| exclusive, single process | 158.2 |

So an exclusive run does about 1.52x the work in the same wall clock. The
practical consequence is that the objectives reported here are pessimistic: with
exclusive CPU each 2400 s run would have done roughly half again as many
evaluations, so the method would be expected to do at least as well and probably
better.

I would rather give you exclusive numbers than caveated ones. The honest problem
is the cost: 20 instances times 5 seeds at 2400 s is about 67 hours
single-process, which I cannot run unattended on this machine without asking
first.

Three options, your call:

1. Rerun everything exclusively at the current 2400 s budget. About 67 hours.
2. Rerun exclusively at 1580 s, which is 2400/1.52 and so matches the evaluation
   count of the current runs rather than the wall clock. About 44 hours, and
   directly comparable to what is in the PR now.
3. Leave the results and document the concurrency and the 1.52x factor in the
   hardware and remarks fields, as I did for the portfolio submission.

I lean towards 2, since it isolates the thing you actually want to compare. Tell
me which and I will run it.

mlschicker: Please do 2. 
I prefer more even number as a time limit like 1800 s for example. 

The problem with benchmarks like this is that we don't know which of the runs were done on efficiency cores, which on performance cores, how much they interfered with each other, ...

Thank you!

PR #51, opened 2026-08-24
[07-independentset] QAOA result for farm instance

AndrewJohnMcD:
Method: QAOA with COBYLA optimization, depths p=1,2,3
Result: MIS = 10 (optimal, approximation ratio 1.0)
Platform: Qiskit AerSimulator, 8192 shots
Runtime: 158.81 seconds (wall-clock, CPU simulator)
Reference: https://quantum.optimisingperformance.com.au/


mlschicker:
Hi @AndrewJohnMcD , 

Would it be possible to run more instances so that we can get a better idea of the performance of you method? 

Moreover, it would be nice, if you generated the README in the canonical format. 
Additional information can always be added if you want. 

You mention that the coefficient are continuous between -2 and 1. 
If you use the canonical QUBO model, they are integer, exactly -2 or 1. 
Do you use a different model?

If you have a single solution, there is no need for a solutions folder, although you can leave it as is. 

Thank you!

Best, 
Max

AndrewJohnMcD:
Thank you to the reviewers for the constructive feedback on this initial submission. The guidance on expanding to multiple instances, adopting the canonical README format, and clarifying coefficient types was invaluable.

A revised and expanded submission addressing all feedback points has been opened as PR #57:
https://github.com/ZIB-AOPT/QOBLIB/pull/57

Key improvements in the new submission:
- **Two instances** (farm, mammalia-kangaroo-interactions) instead of one
- **Canonical README format** following the Egger template
- **Explicit integer coefficients** documented consistently across README and CSV
- **Validated** against `check_submission.py` locally (2/2 passing)

Closing this PR in favour of the revised submission. Thank you again for the guidance.

PR #52, opened 2026-08-28
MarketSplit Quantum Hybrid Submission

VDhriti:
This pull request contributes quantum-hardware results for 16 market-split instances,
comprising four instances at each n = 60, 70, 80, and 90, and D=200. All quantum read-outs were
obtained on `ibm_aachen` (IBM Heron r3). Every submitted solution is feasible, satisfying
`Ax = b` exactly with objective value 0. Each instance directory contains a solution file,
a summary CSV, and a generated README.

## Method

The workflow consists of lattice pre-conditioning, followed by a variational quantum
sampling circuit, with classical post-processing.

## Metadata notes

The underlying research is unpublished. The `Reference` field is therefore  `N/A`, and
implementation-level detail has been limited accordingly; a reference will be provided once
one is available.

Runtimes exclude queue time, in accordance with `CONTRIBUTING.md`. The accounting method is
documented in the submission README.

## Validation

`qoblib-check-submission` reports 16 of 16 instances passing.


mlschicker:
Hi @VDhriti , 

The results look very good. 

As your algorithm is labelled as `Stochastic`, we require at least `5` runs per instance. 

Do you know when the reference will be available? 
When it is, please add it via PR to this submission. 

Thank you!

Best, 
Max

VDhriti:
Hi Max,
Ran the algorithm with 4 more seeds for all of the 16 instances (two stochastic mechanisms in the solver, the warm start and the optimizer). Updated the summary to include the # feasible runs out of 5 and the averaged time across all successful runs. Validation ran clean.
I'm not sure if I can give a date for the reference at this point, but we will be releasing it as soon as we have the ablations and improvements on it done.

Thanks,
Dhriti

01-marketsplit/submissions/20260826_QHybrid_Verma/ms_07_200_248/README.md
```diff
@@ -0,0 +1,42 @@
...
+| Optimality Bound | 0.0 |
+| ====== |  |
+| Modeling Approach | Integer lattice reformulation of Ax=b. |
+| # Decision Variables | 60 |
+| # Binary Variables | 60 |
+| # Integer Variables | N/A |
+| # Continuous Variables | N/A |
+| # Non-Zero Coefficients | 418 |
+| Coefficients Type | Integer |
+| Coefficients Range | 1 - 200 |
```
mlschicker, line 23: For the quantum part, do you use a QUBO formulation? 
If yes, please indicate the respective coefficient range here.

VDhriti: The quantum part does not use a QUBO or Ising formulation. The objective is evaluated classically on the measured bitstrings and is not a quadratic form. This is why we reported the coefficient range as the coefficients of A.

VDhriti:
Hi Max,
Thanks for the review, our circuit is designed in a way that it does not use a QUBO formulation, it uses a classical objective at every iteration. I have left the coefficient range as is.

Let me know if anything else is needed, thanks!

PR #53, opened 2026-08-29
Rename MIP baseline

mlschicker:
- Rename MIP baselines. 
- Change workflow to accept `none` as incumbent.


mlschicker:
The missing solutions in these submissions are a known bug.

Issue #55, opened 2026-09-01
04-steiner: stp_s030_l3_t5_h1_rs24098 is published .opt at 476, but a checker-valid solution exists at 475


marcinjkaminski:
## Summary

`04-steiner/solutions/stp_s030_l3_t5_h1_rs24098.opt.sol` is published at **476** with status
`optimal`. We hold a solution that **your own checker accepts at 475**.

We are reporting this as a question about the label, not as a competitive claim. It is **not**
counted as a beat in our submission, and we have deliberately **withheld this instance** from that
submission so that nothing about your `solutions/` folder changes automatically while you look at
it (see "Why this is an issue and not a row in a PR" below).

## What your checker says

Built from `04-steiner/check` at `2b400f43`, unmodified:

```
$ check_steiner --arcs instances/stp_s030_l3_t5_h1_rs24098/arcs.dat \
                --terms instances/stp_s030_l3_t5_h1_rs24098/terms.dat \
                --sol   ours.sol
Total Cost 475
VALID: Solution successfully verified          (exit 0)

$ check_steiner ... --sol solutions/stp_s030_l3_t5_h1_rs24098.opt.sol
Total Cost 476
VALID: Solution successfully verified          (exit 0)
```

As a non-vacuity control, the same binary rejects a damaged copy of our file:

```
$ check_steiner ... --sol ours_with_one_edge_removed.sol
Terminal node 882 of network 11 not present in its subgraph.
INFEASIBLE: Valid solution file, but it violates the constraints   (exit 21)
```

## Why "the checker is looser than the model" does not explain it

This was our own first hypothesis, and we think it is the one worth ruling out explicitly. The
checker never opens `roots.dat` and does not test acyclicity, so passing it is genuinely weaker
than satisfying `models/integer_linear/stp_node_disjoint.zpl` — the rooted multicommodity-flow ILP
that `models/integer_linear/README.md` identifies as Section 2.2 of Hoang & Koch, *Steiner Tree
Packing Revisited*. A label could in principle be optimal for that model while a checker-valid
solution beat it.

So we did not test our solution against the checker alone. We **constructed the ILP point** — every
`x[a,t]` and every `y[a,k]` — and evaluated all ten `subto` blocks of `stp_node_disjoint.zpl`
literally as written:

```
ours    : ZIMPL objective 475 -- FEASIBLE, all 10 subto blocks satisfied
[...]

PR #56, opened 2026-09-01
04-steiner: 4colors Solver submission (189 instances)

marcinjkaminski:
## 04-steiner — 4colors Solver submission, 189 instances

Heuristic solutions for 189 of the 190 instances of 04-steiner, produced by a classical
Pathfinder-style negotiated-congestion router with exact Dreyfus–Wagner repair. Directory:
`04-steiner/submissions/20260901_4colors-Solver_4colors_v5/`.

**This is a classical result. There is no quantum contribution to any objective value in this
submission.** No QPU, GPU, annealer or quantum simulator was in the loop; `Paradigm` is `Classical`
on every row and `QPU Runtime` is `0` because there was no QPU, not because it was unreported.

### What is claimed

- **189/189 feasible** under `04-steiner/check`, built unmodified from this repo at `2b400f43` and
  run against the submitted `.sol` files as part of building the directory. Total edge count
  **181,057**, with 0 declared-vs-computed mismatches.
- **147 of the 189 have no published reference.** Those values are **first-known**, not beats.
- **Of the 42 that carry a published reference, one is a strict beat**
  (`stp_s070_l4_t6_h0_rs97531`, `.bst` 375 → 374). The other **41 are matched at equal cost, and
  none is worse than its published value.** Of those 41, one is matched by an edge-identical
  packing on a near-forced 4-edge instance; the rest are matched by a structurally different
  packing, verified by arc-set comparison rather than by cost.
- The 147 and the 42 are **never summed** into a combined "beats" figure, here or in the directory
  README.
- `misc/ci/check_submission.py --strict-problem-match` reports **189/189 passed, 0 failed**.

### One instance is deliberately not here — see #55

`stp_s030_l3_t5_h1_rs24098` is published as `.opt` at 476, and we hold a solution your checker
accepts at **475**. We have **withheld it from this submission** rather than let a merge act on it.

The reason is mechanical: `misc/ci/update_bkv.py` unlinks `{instance}.opt.*` before copying in an
improving submitted solution, so shipping our 475 here would **delete a curated proven-optimal
record and downgrade the row to `best known` automatically**, before anyone had ruled on the label.
That is your call to make, not a side effect of merging our PR.

The evidence is in **#55**, including the literal evaluation of our packing against all ten `subto`
blocks of `models/integer_linear/stp_node_disjoint.zpl` and the control that the same evaluator
accepts 29 of 29 of your published `.opt` solutions. We are not claiming that row as a beat and are
not asking for credit on it.

**Verified by dry-run:** with only this submission staged, `update_bkv.py` reports 148 rows newly
attributed to it (147 first-known plus the one `.bst` improvement), the single improvement over an
existing reference is `stp_s070_l4_t6_h0_rs97531: 375 → 374`, and **no `.opt` file is touched**.

### Dual bound

An exact per-net Dreyfus–Wagner relaxation gives a valid lower bound of **177,614** over the
instances submitted here, leaving slack of **3,443 — 1.90%**.

The relaxation drops disjointness *between* trees but keeps the one consequence of it that is free:
a terminal of net *j* lies in net *j*'s tree in every feasible packing, so node-disjoint routing
forbids that node to every other net. Masking each net's foreign terminals out of its own
Dreyfus–Wagner solve is therefore still a relaxation and never weaker — worth 932 units and 647
improved nets over the unmasked version across the full benchmark.

Controls on the bound: it **equals** your published proven optimum on 10 instances; the masked
bound is `>=` the unmasked bound on every net, asserted per net at run time; it agrees with an
independent shortest-path computation on every 2-terminal net; and the falsifier `cost < bound`
fires on 0 of 189.

Most of the remaining slack is bound looseness rather than distance to the optimum: on the
instances where the true answer is published, **99.5% of it is the bound being loose**.

### Optimality claims

`Optimality Bound` is set equal to the objective on **10 rows** — exactly those where the bound
above meets our cost, so `LB == cost == OPT` and the optimum is proven. It is `N/A` on the other
179: we assert no proven optimum outside those 10. Those 10 are all small `s003`/`s004` instances;
our bound does not close on any large instance, which is a limitation of the bound and we would
rather state it than let it be inferred.


marcinjkaminski:
Withdrawing this for internal review before it takes up maintainer time. The submission content itself is unchanged and verified; we will resubmit once we have settled the packaging on our side. Apologies for the noise.

PR #60, opened 2026-09-08
10-topology: improved solutions for three open instances, two proven optimal

marcinjkaminski:
## Summary

Improved solutions for three open small instances of the Order/Degree problem, two of them optimal.

| instance | current | submitted | status |
|---|---:|---:|---|
| `topology_30_6` | 3 | **2** | proven optimal |
| `topology_35_5` | 4 | **3** | proven optimal |
| `topology_50_4` | 5 | **4** | improved, optimality not established |

The class also contains ten large instances taken from the Graph Golf 2019 rankings, all recorded as
`best known`. **For five of them the recorded value is optimal**: it meets the counting bound
exactly, so no graph of that degree can have a smaller diameter at that order. The argument is
purely combinatorial and needs no search. These five rows are therefore closed, and the remaining
five carry a gap of 1 or 2 against the bound.

| instance | n | d | current | counting bound | gap |
|---|---:|---:|---:|---:|---:|
| `topology_512_4` | 512 | 4 | 6 | **6** | 0 |
| `topology_1726_30` | 1726 | 30 | 3 | **3** | 0 |
| `topology_4855_15` | 4855 | 15 | 4 | **4** | 0 |
| `topology_1000000_16` | 1000000 | 16 | 6 | **6** | 0 |
| `topology_1000000_32` | 1000000 | 32 | 5 | **5** | 0 |
| `topology_512_6` | 512 | 6 | 5 | 4 | 1 |
| `topology_1024_4` | 1024 | 4 | 7 | 6 | 1 |
| `topology_9344_6` | 9344 | 6 | 7 | 6 | 1 |
| `topology_100000_8` | 100000 | 8 | 7 | 6 | 1 |
| `topology_65536_6` | 65536 | 6 | 9 | 7 | 2 |

`topology_1000000_32` is the tightest: the bound admits 985,089 vertices at diameter 4 against the
required 1,000,000, a margin of 1.5%.

## The counting bound

A graph of maximum degree `d` and diameter `k` has at most

    1 + d * sum_{i<k} (d-1)^i

vertices, because a breadth-first tree rooted anywhere branches at most `d` ways at the root and at
most `d-1` ways thereafter. An instance `(n, d)` therefore has diameter at least the smallest `k`
for which that expression reaches `n`. Applied to `topology_30_6`, the bound at `k = 1` gives 7 < 30,
so diameter 2 is optimal; applied to `topology_35_5`, the bound at `k = 2` gives 26 < 35, so
diameter 3 is optimal. The five large instances above are the same argument at the `k` one below
their recorded value. `moore_bound.py` in this submission reproduces the whole table from
`instances/`.

For `topology_50_4` the bound gives 53 >= 50 at `k = 3` and does not settle the instance;
`Optimality Bound` is `N/A` there.

## Method

Integer diameter is a plateau objective, so it is not minimized directly. The solver fixes a target
diameter `k` and minimizes a distance-excess objective over degree-preserving moves, under
simulated annealing followed by a deterministic polish. All-pairs distances are computed exactly at
every evaluation, so the reported diameter of any candidate is never an estimate.

## Verification

Every submitted graph was checked with `10-topology/check` built from this tree, and independently
with a second implementation that recomputes exact all-pairs distances and the degree sequence. The
same checker reproduces the recorded diameter of each shipped reference graph, which is how the
implementation was validated before use.


mlschicker:
Hi @marcinjkaminski , 

Please add the instance specific READMEs using the submission checker. 

Thank you!

Best, 
Max

marcinjkaminski:
Hi Max,

Done - the per-instance READMEs are in, generated with
`check_submission.py --generate-readme`.

Thanks for the quick look.

Best,
Marcin

mlschicker:
Hi @marcinjkaminski , 

would it be possible to run the remaining instances so that we gain a better idea of the scaling?
(We don't only want to track the best solution but - after it is reached - the fastest method to achieve it)

Best, 
Max

PR #62, opened 2026-09-08
04-steiner: first values for 147 instances, and two below the recorded values

marcinjkaminski:
## Summary

This submission carries **149** instances of the node-disjoint Steiner tree packing problem:
**147** that have no previously recorded value, and **2** below the recorded one.

## Method

Negotiated-congestion routing, followed by rip-up and reroute, local search and multistart.

The step responsible for the improvements replaces the approximate per-net tree construction with an
exact one on a restricted subgraph, seeded with the approximate tree so the result can never be
worse. The standard construction is a constant-factor approximation, so a polish step reporting no
improvement can still be leaving cost on the table.

## The two improvements

`stp_s070_l4_t6_h0_rs97531` improves from 375 to **374**, reached once in 19 runs.

`stp_s030_l3_t5_h1_rs24098` improves from 476 to **475**. The submitted packing is accepted by the
official checker at cost 475. This row is currently recorded as `optimal` at 476, so the improvement
implies that either the recorded optimum or the solution behind it needs revisiting; the row's
`Remarks` field flags this.

The observation does not depend on this particular packing. Any packing the checker accepts at cost
C yields a feasible point of the shipped rooted ILP with objective at most C, because roots appear
as `terms.dat` entries in all 190 instances, so each net's subgraph can be oriented away from its
own root.

One consequence to flag rather than leave to be discovered: this row is currently backed by
`04-steiner/solutions/stp_s030_l3_t5_h1_rs24098.opt.sol`. Because the submitted value is lower and
asserts no bound of its own, merging will have `update_bkv.py` replace that file with a `.bst.sol`
and record the row as `best known` rather than `optimal`. We raise it rather than assume it is
wanted.

`Optimality Bound` is `N/A` on both.


mlschicker:
Hi @marcinjkaminski , 

it seems like you put the submissions for the individual instances in the submissions directory directly instead of bundling them into their own directory. 

After this change, we can rerun the submission checker. 

Please also add the instance READMEs in this submission. 
This can be done automatically using the submission checker. 

Best, 
Max

marcinjkaminski:
Hi Max,

You are right, and the mistake was mine. `04-steiner/submissions/` did not exist yet, so my copy
made that directory a copy of the submission rather than nesting inside it, which is why the
instance directories ended up one level too high. The other classes already had the directory, which
is why only this one came out wrong.

Force-pushed: everything now sits under `04-steiner/submissions/20260908_4colors-Optimizer_4colors/`,
and the per-instance READMEs are in as well. Ready for another run of the checker whenever suits.

Best,
Marcin

04-steiner/submissions/20260908_4colors-Optimizer_4colors/stp_s030_l3_t5_h1_rs24098/README.md
```diff
@@ -0,0 +1,42 @@
...
+| Algorithm Type | Stochastic |
+| Paradigm | Classical |
+| # Runs | 45 |
+| # Feasible Runs | 45 |
+| # Successful Runs | 10 |
+| Success Threshold | N/A |
+| ====== |  |
+| Hardware Specifications | Intel Core Ultra 9 275HX, 24 cores (8 performance + 16 efficiency), 93 GB RAM, Ubuntu 24.04 |
+| ====== |  |
+| Total Runtime | 2507.678 |
```
mlschicker, line 35: As there is no optimality bound provided, I understand that this method is a heuristic. 
What are the stopping criteria?
Asking because of the big difference between runtime and TTS

04-steiner/submissions/20260908_4colors-Optimizer_4colors/stp_s030_l3_t5_h1_rs24098/README.md
```diff
@@ -0,0 +1,42 @@
...
+| Time to Solution | 188.470 |
+| CPU Runtime | 2507.678 |
+| GPU Runtime | N/A |
+| QPU Runtime | N/A |
+| Other HW Runtime | N/A |
+| ====== |  |
+| Remarks | This instance is currently recorded as optimal at 476; the packing submitted here is accepted by the official checker at 475, so the recorded value merits review. Classical result: no quantum resource was used. |
```
mlschicker, line 42: I will check this. 
Referring to #55

PR #63, opened 2026-09-08
03-birkhoff: 269 improved solutions, 15 first values, 18 proven optimal

marcinjkaminski:
## Summary

This submission carries **284** instances of the minimum Birkhoff decomposition problem that improve
the best-known value, out of the 375 in the library.

| | instances | reduction |
|---|---:|---:|
| Lower than the current value | 269 | 9,771 |
| -- dense | 130 | 2,310 |
| -- sparse | 139 | 7,461 |
| First value (no previous entry), all `n = 64` and `n = 100` dense | 15 | -- |
| **total** | **284** | **9,771** |

Optimality is asserted on the **18** submitted instances where the enumeration closed.

The enumeration also settles 47 further instances at their recorded values, confirming 37 rows
already marked `optimal` and establishing optimality for 10 more that are currently recorded
`best known`. Those ten are reported separately as a status change.

The current value is held by the Birkhoff+ submission (Valls, IBM Research Europe) on 267 of these
instances, by a Gurobi baseline on one, and by the E-FCFW submission on one.

The pattern by size is uniform. Below `n = 7` the existing values are already optimal or nearly so
and we match them. From `n = 7` upward every family improves, with the margin widening:

| n | kind | lower | equal | reduction |
|---:|---|---:|---:|---:|
| 7-9 | dense | 29 | 1 | 115 |
| 10-16 | dense | 70 | 0 | 666 |
| 24-48 | dense | 30 | 0 | 1,528 |
| 7-9 | sparse | 20 | 10 | 119 |
| 10-16 | sparse | 69 | 1 | 2,089 |
| 24-100 | sparse | 50 | 0 | 5,259 |

The largest single reductions are at `n = 100` dense, where no value was previously recorded, and at
`n = 100` sparse: `B100_100_4` from 493 to **276**, `B100_100_5` from 492 to **277**,
`B100_100_2` from 487 to **272**.

Values are reported over 4-6 independent trials per instance differing only in random seed. The
returned value was identical across all trials on 216 of the 375 instances.

## Method

Four components run under a per-instance time budget: a greedy bottleneck construction; a
minimum-support integer program over an enumerated pool of candidate permutations; a
large-neighborhood search whose subproblems are small enough that every legal column can be
enumerated, so each move is exact; and a structural refinement that uses the fact that every row of
the matrix induces a partition of the same weight multiset. All arithmetic is on the scaled
integers.

## Optimality

Asserted on 18 submitted instances, all with `n <= 9`. Every term of a valid decomposition is a perfect
matching of the instance support, so at these sizes the candidate set is finite and can be
enumerated in full; the minimum-support program then returns infeasible at `k-1`.

Two independent checks accompany the claim, both reproducible from `PROOF.md`:

- **Enumeration completeness.** The number of enumerated permutations is compared against Ryser's
  formula for the permanent of the 0/1 support matrix, on all 65 instances where the enumeration
  closed -- the 18 asserted here and the 47 further instances. No discrepancies.
- **Solver-free replication at n = 3.** The infeasibility results are reproduced by exhaustive
  enumeration over every candidate subset in exact rational arithmetic, with no solver involved.

For `n >= 4` the infeasibility result comes from a MIP solver and is contingent on its correctness.
We state this because the class checkers do not test optimality, so the assertion is not
independently validated by CI.

## Scope of the sparse results

The instance files contain, alongside the matrix, the `weights` and `permutations` from which it was
generated. On sparse instances those fields constitute a valid decomposition of size `n`, and on all
139 sparse instances where our value is lower than the current one, that decomposition is lower
still. The values submitted here were obtained from the matrix alone and do not read those fields.
On 44 instances, chiefly `n >= 48`, our value exceeds the size of that decomposition, by 4,087 terms
in aggregate.

For dense instances the corresponding decomposition has `n^2` terms -- 10,000 at `n = 100` against
current values near 490 -- and does not bear on those results.

What this implies for the recorded best-known values is raised separately as an issue, together with
a full set of decompositions derived from those fields.

## Verification

All 375 solutions produced, submitted or not, were checked with `03-birkhoff/check` (v1.2) built
from this tree, and
independently with a verifier in exact integer arithmetic that recomputes the reconstruction and
the convexity condition. The two agree on every instance. Weights are emitted as integers scaled by
the instance scale, so no floating-point tolerance is involved on our side.


mlschicker:
Hi @marcinjkaminski , 

thank you for the submission. 

Would it be possible to provide an objective time series?
(This information is interesting for all of the submission you just made)

Best, 
Max

PR #64, opened 2026-09-08
06-portfolio: 8 improved solutions and first values for 224 instances, 96 proven optimal

marcinjkaminski:
## Summary

**232** solutions across the a003, a004, a005, a050, a200 and a400 families.

| | instances |
|---|---:|
| Lower than the current value | 8 (all a050) |
| First value (no previous entry), proven optimal | 96 (a003, a004, a005) |
| First value, heuristic | 128 (a200, a400) |
| **total** | **232** |

Optimality is asserted on the **96** first values, where the dynamic program is exhaustive.

The same program settles the 64 a010 instances at their recorded values, of which 50 are currently
recorded `best known`. That is a status change rather than a value change and is reported
separately, with the proof code attached.

The eight improvements are all at high risk aversion, where the one-hour reference is weakest.

| instance | current | submitted | delta |
|---|---:|---:|---:|
| `a050_t10_orig_b020_l1e-02` | -37,107 | **-40,774** | -3,667 |
| `a050_t15_s00_b020_l1e-02` | -43,792 | **-46,238** | -2,446 |
| `a050_t10_s01_b020_l1e-02` | -2,248 | **-4,026** | -1,778 |
| `a050_t10_s00_b020_l1e-02` | -12,340 | **-13,546** | -1,206 |
| `a050_t10_s02_b020_l1e-02` | -43,565 | **-44,771** | -1,206 |
| `a050_t15_s02_b020_l1e-03` | -316,799 | **-316,930** | -131 |
| `a050_t15_orig_b020_l5e-04` | -397,497 | **-397,568** | -71 |
| `a050_t10_s00_b020_l5e-04` | -234,259 | **-234,277** | -18 |

## Method

The reference model reduces exactly to a chain over periods with state given by the unit-count
vector. The shipped formulation carries three interchangeable unit copies per (asset, direction,
period), producing a symmetric block per group that weakens the dual bound; working in counts
removes the symmetry without changing the feasible set. For the small budgets the chain is solved
exactly by dynamic programming. For a050 the budget puts the state space beyond exhaustive
treatment, and an iterated local search in count space is used, warm-started from the reference
solution.

## Optimality

Asserted on 96 instances, where the dynamic program enumerates every feasible position plan and
therefore returns the global optimum of the reference model.

The implementation is a standalone C++17 program with no external dependencies, included as
`portfolio_dp.cpp` so the result can be reproduced directly from the instance files. It shares no
code with our heuristics. It was validated three ways before the claim was made:

- its objective function reproduces the official checker's value on all 296 solutions produced,
  submitted or not;
- its optimum agrees with exhaustive enumeration over all plans on the 96 small instances;
- the plans it emits are accepted by the official checker at the stated objective on all 160
  instances where optimality was established, submitted or not.

`Optimality Bound` is `N/A` on the eight a050 improvements and on the 128 a200 and a400 first
values, which are heuristic.

## Instance keys

Directory names, the `Problem` column and the solution filenames all use the canonical instance key
- no `po_` prefix, lambda exponent zero-padded, e.g. `a050_t10_orig_b020_l1e-02`. That is the form
`misc/ci/update_bkv.py` derives from the CSV before it looks for the solution file, so naming them
this way is what lets the file be found. Existing submissions in this class use the `po_`-prefixed
form; happy to switch if you would rather they were consistent.

## Parameters

All objectives are evaluated at the checker defaults: capital `C = 10` (cash 1e6, unit 1e5),
`ub = 3`, `cs1 = 4`, `cs2 = 7`. This is stated explicitly because `instances/manifest.json` records
the budget and lambda per instance but not the cash and unit values, so the same solution scores
differently under a different capital.


mlschicker:
Hi @marcinjkaminski , 

thank you for the submission. 

Please do not upload the code directly in the submission. 
Instead, create a public GitHub Repo on your account and link to it in the reference. 

Thank you!

Best, 
Max

PR #70, opened 2026-09-12
Add four verified Market Split submissions (Kenji R5 + Aer)

taiyonoukou-commits:
### Change type

- [x] New submission package
- [ ] Update to an existing submission package
- [ ] Submission fixes (format/validation only)

### Scope

- **Problem class:** `01-marketsplit`
- **Submission root:** `01-marketsplit/submissions/20260913_KenjiR5_Aer_Kanamori`
- **Instances:** `ms_04_050_001`, `ms_04_050_004`, `ms_04_050_005`, `ms_04_100_009`

### Contents

Each instance contains its generated README, objective time series, five-run evidence, solution, and summary CSV. All four submissions report five actual runs.

### Validation

- [x] Followed `CONTRIBUTING.md`
- [x] Ran the authoritative local submission checker with full solution validation
- [x] Result: **4/4 instances passed, 0 failed — Overall: OK**
- [x] Every recorded run passed the fixed post-measurement estimator and official validator
- [x] No classical final repair was used

### Execution and safety

The runs used Qiskit Aer on CPU as a quantum simulator. No IBM account, IBM credentials, physical QPU access, or QPU submission was used. No Kenji/UJ implementation source is included.

This is a separate new submission root and does not modify PR #69 or its files.

PR #69, opened 2026-09-11
Add Kenji R5 Qiskit Aer results for Market Split and Sports

taiyonoukou-commits:
## Summary

Submission by **Kenji Kanamori** (Independent Researcher).

- 19 Market Split instances, 5 independent Qiskit Aer runs each (95 runs)
- 1 Sports / ITC2021 instance, 5 independent Qiskit Aer runs
- 100 independent simulator runs across 20 instances
- Fixed Kenji R5 + Universal Joint measurement-estimation pipeline

## Results and validation

- Market Split: 95/95 feasible runs; official checker PASS for all runs
- New N20 rollout: 12 instances and 60/60 PASS
- N20 execution uses 10+10 quantum shards with interface reconstruction
- Sports ITC2021_Early_14: 5/5 hard-feasible RobinX validations
- Sports soft objectives: 3139, 2569, 2571, 3218, 3028; best run 2569 submitted
- SHA-256 manifests and five-run operational evidence are included

## Method and disclosure boundary

The Kenji R5 estimator is a proprietary black box and its source is not included.
Its fixed authority SHA-256 is
`958AC865431968359C3318815C198A228350982FF8D489E6BC252B57E0E3E156`.

Only measurement-derived released outputs and non-secret execution metadata are
included. Reference solutions and validator outcomes were not available to the
estimator before release. No classical solver or classical final repair produced
the submitted answers.

The N20 rollout first measured the direct route, then replaced it with smaller
10-qubit shared shard circuits after the direct route showed insufficient candidate
yield. The submitted route achieved 60/60 independent-run validation passes.

IBM hardware results with fewer than five independent runs per instance were
deliberately excluded from this PR.

PR #73, opened 2026-09-14
[01-marketsplit] Reorganize 23 Kenji R5 Aer submissions

taiyonoukou-commits:
## Scope

Maintainer-requested reorganization of the existing results in PRs #69 and #70 into one problem-class PR and one submission directory.

- Problem class: Market Split
- Submission root: `01-marketsplit/submissions/20260915_KenjiR5_Aer_Kanamori`
- Instances: 23 (19 from #69, 4 from #70)
- Files: 118

## Integrity

- All 118 files are Git-blob identical to their source PR artifacts.
- Every instance retains its solution, README, summary, run evidence, and objective time series.
- Solution contents and SHA-256 values match the source PRs.
- No truth/reference/known-solution material was introduced.
- No Kenji, Universal Joint, checker, workflow, or other implementation file was changed.
- No recomputation or new answer generation was performed. `RUN_QPU=false` was maintained; no IBM/provider/API/QPU call was made.

## Validation

- QOBLIB authoritative local submission checker: PASS (23/23 instances)
- Independent reorganization audit: PASS (one submission root; no duplicate/missing instance; all required files present; all source blobs match)

## Preserved source note

The `SHA256SUMS.txt` file inherited from #69 is preserved byte-for-byte as requested. Its solution/evidence/summary entries verify. Seven README entries were already stale in #69 after README regeneration; the same seven source mismatches are therefore retained rather than rewriting an existing artifact.

The source PRs remain open pending independent review of the replacement PRs.


taiyonoukou-commits:
Hi Max,

One additional point regarding reproducibility for PRs #73–#76:

If useful, I would also be happy to perform a live blinded reproducibility test by screen sharing.

The QOBLIB team could prepare and provide the quantum-computation Python code and previously unseen test problems during the live session, without disclosing the reference answers to me.

I would first perform a limited safety check of the provided code to confirm that it does not access or probe Kenji's proprietary black-box implementation or perform unrelated external actions. If clear, I would run the provided code in Google Colab, changing only the path needed to connect it to the fixed Kenji implementation identified by its SHA-256, and return Kenji's outputs live.

The reference answers would remain known only to the QOBLIB side until after Kenji has produced its results.

To protect the black-box implementation from query-based probing, I would propose a limit of up to 10 previously unseen problems, with one Kenji evaluation per problem.

This would allow the behavior of the fixed Kenji implementation to be independently observed on previously unseen inputs without disclosing its internal implementation.

Best,
Kenji Kanamori

